PR #36, opened 2025-11-24
Generate top-level and individual API reference docs for all SDK clients

arandito:
### Description

This PR adds documentation generation for all AWS clients in the `aws-sdk-python` repo. Users can either generate a single top-level documentation site for all clients or independently for individual clients (see Testing).

Documentation is built using:
- [Material for Mkdocs](https://squidfunk.github.io/mkdocs-material/getting-started/) - The Markdown theme to generate the static html files for the documentation site
- [mkdocstrings](https://mkdocstrings.github.io/) - Auto-documents python modules using Markdown docstrings
- [griffe](https://mkdocstrings.github.io/griffe/) - Parses client packages to generate stub files used by mkdocstrings to generate documentation

### Testing
For top-level client docs, execute these commands in the root level of this repo. For individual client documentation, execute these commands in each client's directory.
```bash
# Install docs dependencies
make docs-install

# Build static HTML to site/
make docs

# Serve documentation locally to port 8000
make docs-serve

# Clean up artifacts
make docs-clean
```


mkdocs.yml
```diff
@@ -0,0 +1,82 @@
+site_name: AWS SDK for Python
+site_description: Documentation for AWS SDK for Python Clients
+
+copyright: Copyright &copy; 2025, Amazon Web Services, Inc
```
jonathan343: We should look into generating the Copyright year dynamically so we don't have to remember to update it manually. Doing some quick research, [this discussion](https://github.com/squidfunk/mkdocs-material/discussions/4969) has a few options that all seem reasonable:
1. **[preferred]** Use the [hooks](https://www.mkdocs.org/user-guide/configuration/#hooks) system to create a plugin that will format the copyright string. Similar to [this](https://github.com/squidfunk/mkdocs-material/discussions/4969#discussioncomment-13027051)
2. Inject something like `<script>document.write(new Date().getFullYear())</script>` directly into copyright config value in `mkdocs.yml`

mkdocs.yml
```diff
@@ -0,0 +1,82 @@
...
+repo_name: awslabs/aws-sdk-python
+repo_url: https://github.com/awslabs/aws-sdk-python
+
+theme:
+  name: material
```
jonathan343, line 11: I think we can improve branding by adding the AWS logo. This would involve adding the logo to an `docs/assets` folder and then referencing that somewhere here like:
```
logo: assets/aws-logo-white.png
```

You can see my original [POC](https://jonathan343.github.io/aws-sdk-python/) for how this would look.

**Note:** I'm not sure if it's possible for us to easily add the white logo for dark mode and black logo for light more. In my POC I needed to keep the navbar one color for both light and dark more. But you might want to try and figure that out.

mkdocs.yml
```diff
@@ -0,0 +1,82 @@
...
+  palette:
+    - scheme: default
```
jonathan343: So even if my OS is set to dark mode, we will render the light more page for users? If so, that doesn't feel like a good user experience. We should consider adding the automatic mode:

Check this out: https://github.com/jonathan343/aws-sdk-python/blob/docs-poc/mkdocs.yml#L17-L38

docs/stylesheets/extra.css
```diff
@@ -0,0 +1,3 @@
+.md-grid {
```
jonathan343: nit - More of a personal opinion, but I'm not a big fan of this. I like the default used by mkdocs/material and  think it's best to stick with them unless there is a really good reason not to. We can see what others think about it.

pyproject.toml
jonathan343, line 1: This is `pyproject.toml` is conflicting with the meta package in `clients/aws-sdk-python`. We shouldn't define separate files with the same package name. I'm thinking we might actually be able move `clients/aws-sdk-python` to the top-level of this repo. That might actually make more sense. Will need some looking into

pyproject.toml
```diff
@@ -0,0 +1,20 @@
+[project]
+name = "aws-sdk-python"
+version = "0.1.0"
+description = "Add your description here"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
+    "mkdocs-awesome-pages-plugin>=2.10.1",
+]
```
jonathan343: Flagging this so we don't forget about it. If we end up needing this, it should be removed from both `dependencies` and the `dev` dependency group. It should just get added with the others in `docs`

jonathan343 (commented):
Thanks Antonio, this is a great start. I've left some initial comments around the configuration of the documentation. I'll get to the generation scripts later tonight, but wanted to comment on what I've seen so far.

Additionally, I have some high-level comment around the top-level clients page:

Image

1. We need to remove the `None` links and replace them with individual client links.
1. The individual client links throw `404 - Not found`.
1. We need to avoid including folders like `.ruff_cache`

mkdocs.yml
jonathan343, line 1: nit - We should add something like [this]([github/gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore#L167-L168)) to our `.gitignore` file:

```
# mkdocs documentation
/site
```

CONTRIBUTING.md
```diff
@@ -1,3 +1,4 @@
+<!-- --8<-- [start:docs] -->
```
jonathan343: Why do we need the start/end indicators? IS there something other than the relative link that is requiring this? We might just want to update that link instead of working around it

mkdocs.yml
```diff
@@ -0,0 +1,82 @@
...
+            filters:
+            - "!^_"
+            - "!^deserialize"
+            - "!^serialize"
+
+markdown_extensions:
+  - pymdownx.highlight
+  - pymdownx.inlinehilite
+  - pymdownx.snippets:
+      base_path: ['.']
```
jonathan343:
```suggestion
      base_path: ['.']
      check_paths: true
```

I think we should set `check_paths: true` which will _"Make the build fail if a snippet can't be found."_

Reference: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#options

scripts/generate_nav.py
```diff
@@ -0,0 +1,141 @@
...
+    """Build navigation structure for clients."""
+    nav_structure = [
+        {
+            "Getting Started": [
+                {"Overview": "index.md"},
+                {"Contributing": "contributing/index.md"},
+            ]
+        },
+        {
+            "Clients API Reference": [
```
jonathan343: nit - "Clients API Reference" sounds a bit awkward. Maybe just "API Reference"
```suggestion
            "API Reference": [
```

scripts/generate_doc_stubs.py
jonathan343, line 1: nit - Maybe nest these under `scripts/docs/...` for better organization. We may add other non-docs scripts in the future

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+Generate API documentation for AWS SDK Python clients.
+
+This script generates MkDocs Material documentation that matches the reference
+documentation format. It analyzes Python client packages using griffe
+and creates structured documentation with proper formatting.
+"""
+
+import logging
+import sys
+from dataclasses import dataclass, field
```
jonathan343: nit - unused import
```suggestion
from dataclasses import dataclass
```

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+        if models.enums:
+            content += "## Enums\n\n"
+            for enum in sorted(models.enums, key=lambda x: x.name):
+                content += f"- [`{enum.name}`](enums/{enum.name}.md)\n\n"
+
+        docs_path = f"clients/{self.service_name}/index.md"
+        with mkdocs_gen_files.open(docs_path, "w") as f:
+            f.write(content)
+
+        logger.info(f"Generated index.md")
```
jonathan343:
```suggestion
        logger.info("Generated index.md")
```

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+
+@dataclass
+class OperationInfo:
+    """Information about a client operation."""
+
+    name: str  # e.g., "converse"
+    module_path: str # e.g., "aws_sdk_bedrock_runtime.client.BedrockRuntimeClient.converse"
+    input: StructureInfo # e.g., "ConverseInput"
+    output: StructureInfo # e.g., "ConverseOperationOutput"
+    stream_type: StreamType.NONE
```
jonathan343: This either needs to be just `StreamType` or `StreamType = StreamType.NONE` if you meant to set a default

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+
+    def _analyze_client_package(self, package_name: str) -> ClientInfo:
+        """Analyze a client package using griffe."""
+        logger.info(f"Analyzing package: {package_name}")
+        package = griffe.load(package_name)
+
+        # Parse submodules
+        client_module = package.modules.get("client")
+        config_module = package.modules.get("config")
+        models_module = package.modules.get("models")
```
jonathan343: nit - Should we add validation to make sure none of these are `None`?

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+
+def main() -> int:
+    """Main entry point for the documentation generator."""
+    repo_root = Path(__file__).parent.parent.absolute()
+    output_dir = repo_root / "docs" / "clients"
+    clients_dir = repo_root / "clients"
+
+    try:
+        for client_dir in clients_dir.iterdir():
+            if client_dir.is_dir() and client_dir.name != "aws-sdk-python":
```
jonathan343: When testing, this was picking up `.ruff_cache`. We should add another check that verifies there is a `pyproject.toml` file or something

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+        .replace("aws-sdk-", "")
+        .replace("-", " ")
+        .title()
+    )
+
+
+def main() -> int:
+    """Main entry point for the documentation generator."""
+    repo_root = Path(__file__).parent.parent.absolute()
+    output_dir = repo_root / "docs" / "clients"
```
jonathan343: Is this still accurate? I see `docs_dir` defined in the constructor of `DocStubGenerator` but don't think it ever gets used.

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+        """Analyze a client package using griffe."""
+        logger.info(f"Analyzing package: {package_name}")
+        package = griffe.load(package_name)
+
+        # Parse submodules
+        client_module = package.modules.get("client")
+        config_module = package.modules.get("config")
+        models_module = package.modules.get("models")
+
+        client_class = self._find_class_with_suffix(client_module, "Client")
```
jonathan343: Unlikely case, but since it's possible, we should guard around a possible `None` return value:
```
if not client_class:
    raise ValueError(f"No class ending with 'Client' found in {package_name}.client")
```

scripts/generate_doc_stubs.py
```diff
@@ -0,0 +1,590 @@
...
+        self._generate_errors(client_info.models.errors)
+        self._generate_unions(client_info.models.unions)
+        self._generate_enums(client_info.models.enums)
+
+    def _generate_index(self, client_info: ClientInfo) -> None:
+        """Generate the main index.md file."""
+        content = f"# {self.service_name}\n\n"
+
+        # Client section
+        content += "## Client\n\n"
```
jonathan343: Rather than doing all these string concatenations, it might be better to have some `lines` list that we append to and then later generate the content with `content = "\n".join(lines)`

scripts/generate_nav.py
```diff
@@ -0,0 +1,141 @@
...
+            ]
+        }
+    ]
+    mkdocs_gen_files.config["nav"] = nav_structure
+    logger.info(f"✅ Generated navigation structure for {len(clients)} clients")
+
+
+def main() -> int:
+    """Main entry point to generate navigation."""
+    repo_root = Path(__file__).parent.parent
```
jonathan343: The other script uses `repo_root = Path(__file__).parent.parent.absolute()` we should try to stay consistent

jonathan343 (commented):
More (mostly minor) comments after reviewing the `generate_doc_stubs.py` script. I know some of the navbar stuff is WIP, so I'll probably wait to review that file

mkdocs.yml
arandito: Good catch, added a hook for this!

mkdocs.yml
arandito: I added white and black logos to our assets folder. It switches automatically with the theme and required a very small css customization.

mkdocs.yml
arandito: Added the automatic mode to default to the system.

docs/stylesheets/extra.css
arandito: I added it initially to expand the center column a bit more but I realized that it is easier to read the documentation if the column isn't very wide. I reverted this.

pyproject.toml
arandito: Removed it!

pyproject.toml
arandito: After offline discussion: 
We are not including these dependencies in the meta package to avoid coupling the package with our top-level documentation process. We are going to use a `requirements-docs.in` for top-level documentation dependencies instead. Each client will have a `docs` dependency group in their `pyproject.toml` to build documentation independently.

mkdocs.yml
arandito: Turns out we didn't have a `.gitignore` file. I created one and added a couple files!

CONTRIBUTING.md
arandito: Didn't realize I could change the link to use the full github.com url. Updated!

mkdocs.yml
arandito: Added `check_paths: true`. Removed `base_path: ['.']` since it is the default.

scripts/generate_doc_stubs.py
arandito: Moved them!

scripts/generate_doc_stubs.py
arandito: Done.

scripts/generate_doc_stubs.py
arandito: Done.

scripts/generate_doc_stubs.py
arandito: Added some validation and logging.

scripts/generate_doc_stubs.py
arandito: Yea this was left over from a previous iteration of the script. Removed!

scripts/generate_doc_stubs.py
arandito: Added!

scripts/generate_nav.py
arandito: Updated to use `repo_root = Path(__file__).parent.parent.absolute()`

scripts/generate_doc_stubs.py
arandito: Good point, switched to use a `lines` list. This actually helps better understand what gets rendered. Blank lines are now intentional `lines.append("")` instead of previously relying on `"\n\n"`.

scripts/generate_doc_stubs.py
arandito: Good catch, this was causing the `None` 404 issue you mentioned in your top-level comment. I fixed this iteration.

scripts/generate_doc_stubs.py
arandito: Yep! Switched to `StreamType | None`

scripts/generate_nav.py
arandito: Fixed

mkdocs.yml
```diff
@@ -0,0 +1,101 @@
+site_name: AWS SDK for Python
+site_description: Documentation for AWS SDK for Python Clients
+
+repo_name: awslabs/aws-sdk-pythons
```
jonathan343:
```suggestion
repo_name: awslabs/aws-sdk-python
```

mkdocs.yml
```diff
@@ -0,0 +1,101 @@
...
+repo_url: https://github.com/awslabs/aws-sdk-python
+
+exclude_docs: |
+  README.md
```
jonathan343: Why is this needed?

mkdocs.yml
```diff
@@ -0,0 +1,101 @@
...
+    - navigation.instant
+    - navigation.top
+    - search.suggest
+    - search.highlight
+    - content.code.copy
+
+plugins:
+  - search
+  - literate-nav:
+      nav_file: SUMMARY.md
```
jonathan343, line 47: Can you explain this SUMMARY.md file and how it's used?

scripts/docs/generate_nav.py
```diff
@@ -0,0 +1,91 @@
...
+            return 1
+    except Exception as e:
+        logger.error(f"Unexpected error: {e}")
+        return 1
+
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
jonathan343: nit - A lot of these fines are missing a newline at the end of the file. We should add them.

For the Python files, we'll need a long-term solution for automating linter/formatter.

For now we can do `uvx ruff check <script>` and `uvx ruff format <script>`. It catches a few formatting issues and handles the newlines for us

Makefile
```diff
@@ -0,0 +1,25 @@
...
+docs-generate:
+	uv run python scripts/docs/generate_all_doc_stubs.py
+	uv run python scripts/docs/generate_nav.py
+
+docs: docs-generate
+	uv run mkdocs build
+
+docs-serve:
+	@[ -d site ] || $(MAKE) docs
+	uv run python -m http.server $(DOCS_PORT) --bind 127.0.0.1 --directory site
```
jonathan343: We may have discussed this before, but why not use something like `uv run mkdocs serve --dev-addr 127.0.0.1:$(DOCS_PORT)`?

clients/aws-sdk-bedrock-runtime/pyproject.toml
```diff
@@ -34,6 +34,12 @@ test = [
     "pytest-asyncio>=0.20.3,<0.21.0"
 ]
 
+docs = [
+    "mkdocs~=1.6.1",
```
jonathan343: We should try to keep constraints consistent between all clients and the top-level requirements

mkdocs.yml
arandito: Fixed!

mkdocs.yml
arandito: This was left over from the individual client `mkdocs.yml`. Each client's `docs/` directory has a `README.md` and a warning gets thrown since it conflicts with `index.md`:
`WARNING -  Excluding 'README.md' from the site because it conflicts with 'index.md'.`

This isn't required in the top-level configuration since there isn't a `README.md` inside of the root `docs` directory. Will remove.

mkdocs.yml
arandito: SUMMARY.md is a file generated by the `generate_nav.py` script. It contains the navigation structure for the site. 

For example:
```
* [Overview](index.md)
* [Contributing](contributing.md)
* [Available Clients](clients/index.md)
    * [Bedrock Runtime](clients/bedrock-runtime/index.md)
    * [Sagemaker Runtime Http2](clients/sagemaker-runtime-http2/index.md)
    * [Transcribe Streaming](clients/transcribe-streaming/index.md)
```
The [`mkdocs-literate-nav`](https://oprypin.github.io/mkdocs-literate-nav/) plugin reads this file and builds the navigation structure as part of the mkdocs build.

This replaces the manual `nav` configuration inside of `mkdocs.yml` and allows us to generate navigation dynamically for all clients. This file gets cleaned up as part of `make docs-clean` so it gets regenerated each build.

scripts/docs/generate_nav.py
arandito: Added trailing newlines to files missing them and ran `uvx ruff check <script>` and `uvx ruff format <script>`.

Makefile
arandito: The issue with `mkdocs serve` is that it forces a rebuild of the entire docs site each time you run it. It doesn't create a `site/` directory like `mkdocs build`. If an engineer accidentally interrupts or exists the server process, running `mkdocs serve` again forces a rebuild. As the documentation grows, this makes rebuilds longer.

This allows us to always serve the `site/` directory if it exists and builds the site if it doesn't.

clients/aws-sdk-bedrock-runtime/pyproject.toml
arandito: This was a miss. Will fix it!

jonathan343 (approved):
Looks good to merge to the feature branch

PR #37, opened 2025-11-24
Add integration tests for Bedrock Runtime

alexgromero:
*Issue #, if available:*

*Description of changes:*
This PR adds comprehensive integration tests for Amazon Bedrock Runtime client covering all streaming patterns:
- Non-streaming operations using `converse`
- Output streaming operations using `converse_stream`
- Bidirectional streaming operations using `invoke_model_with_bidirectional_stream`


clients/aws-sdk-bedrock-runtime/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,305 @@
...
+
+SESSION_END_EVENT = """{
+    "event": {
+        "sessionEnd": {}
+    }
+}"""
+
+
+@pytest.fixture
+async def bedrock_client():
```
jonathan343: nit - I don't think we need to copy/past this fixture from `conftext.py` just to configure a different region. Instead, we can do something like have a `_create_bedrock_client` function that accepts a region as a parameter. Then define separate fixtures for each region:

```py
def _create_bedrock_client(region: str):
    """Helper to create a BedrockRuntimeClient for a given region."""
    http_client = AWSCRTHTTPClient()
    return BedrockRuntimeClient(
        config=Config(
            endpoint_uri=f"https://bedrock-runtime.{region}.amazonaws.com",
            region=region,
            aws_credentials_identity_resolver=create_default_chain(http_client),
            transport=http_client,
        )
    )


@pytest.fixture
async def bedrock_client():
    """Create a BedrockRuntimeClient for us-west-2."""
    return _create_bedrock_client("us-west-2")


@pytest.fixture
async def bedrock_client_us_east_1():
    """Create a BedrockRuntimeClient for us-east-1 (for nova-sonic model)."""
    return _create_bedrock_client("us-east-1")
```

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
```diff
@@ -0,0 +1,64 @@
...
+from aws_sdk_bedrock_runtime.config import Config
+
+AWS_REGION = "us-west-2"
+ENDPOINT_URI = f"https://bedrock-runtime.{AWS_REGION}.amazonaws.com"
+MODEL_ID = "amazon.titan-text-express-v1"
+BIDIRECTIONAL_MODEL_ID = "amazon.nova-sonic-v1:0"
+
+
+@pytest.fixture
+async def bedrock_client():
```
jonathan343: Why does this need to be async?

clients/aws-sdk-bedrock-runtime/tests/integration/test_non_streaming.py
```diff
@@ -0,0 +1,41 @@
...
+    output_message = response.output.value
+    assert output_message.role == "assistant"
+    assert len(output_message.content) > 0
+
+    content_block = output_message.content[0]
+    assert isinstance(content_block, ContentBlockText)
+    assert isinstance(content_block.value, str) and content_block.value
+
+    response_text = content_block.value.lower()
+    assert "guido" in response_text, (
```
jonathan343: I'm curious how consistent you've seen responses from both models. In the past, I've noticed some variance in the responses from the service. Instead of answering the question, it would reply with stuff like:
- `Sorry, this model is unable to browse the internet and therefore is not able to provide any information on Smithy.`
- `Sorry - This model is designed to avoid generating stereotypes. However, it can provide general information on the topic.`
I'm asking because in situations like this, the assertion for `guido` in the text would fail. Maybe it's just because I'm asking about something not as popular as Python (smithy). I'm a bit worried about brittleness

jonathan343 (commented):
Thanks Alessandra, this looks pretty good already. Just have a few questions

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
alexgromero: Good catch! I initially made it async because the client has async methods, but the fixture itself doesn't await anything so I'll update that.

clients/aws-sdk-bedrock-runtime/tests/integration/test_non_streaming.py
alexgromero: #### Running 100 tests with model: `amazon.titan-text-express-v1`:
```
================================================================================
SUMMARY
================================================================================
Total runs: 100
Mentions 'Guido': 99 (99.0%)
No mention: 1 (1.0%)
```
The single miss (Run 26) returned:
  "Sorry - this model is unable to respond to this request."
However, I ran another 300 iterations afterwards with 100% success rate.


#### Running 100 tests with model: `amazon.nova-sonic-v1:0`
```
================================================================================
SUMMARY
================================================================================
Total runs: 100
Mentions 'Guido': 100 (100.0%)
No mention: 0 (0.0%)
```

The one failure appears to be a transient service error rather than an incorrect model response. Given the 99.75% reliability (399/400 successful), I think this assertion is appropriate for integration tests. That said, I'm open to removing or loosening it if you prefer.

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+
+def _create_bedrock_client(region: str):
+    """Helper to create a BedrockRuntimeClient for a given region."""
+    http_client = AWSCRTHTTPClient()
+    return BedrockRuntimeClient(
+        config=Config(
+            endpoint_uri=f"https://bedrock-runtime.{region}.amazonaws.com",
+            region=region,
+            aws_credentials_identity_resolver=create_default_chain(http_client),
+            transport=http_client,
```
nateprewitt: Do we need to supply the HTTP client in this case? Shouldn't it [default to AWSCRTHTTPClient](https://github.com/awslabs/aws-sdk-python/blob/2d8f9746d1fa5add34c8d46e717a45e70b62d90f/clients/aws-sdk-bedrock-runtime/src/aws_sdk_bedrock_runtime/config.py#L200)?

Ideally, we want to try to test the closest setup to what a customer would get out of the box unless we're validating specific behaviors.

**Edit:** We can use `EnvironmenCredentialsProvider` for now if the concern is the http client for the chain. We still don't have that format entirely defined, so I wouldn't suggest we rely on it just yet.

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+        )
+    )
+
+
+@pytest.fixture
+def bedrock_client():
+    """Create a BedrockRuntimeClient for us-west-2."""
+    return _create_bedrock_client("us-west-2")
```
nateprewitt: nit; do we need to make these fixtures or is the helper function sufficient? Setting up a new client per test isn't notable overhead and this may be a premature optimization.

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
```diff
@@ -0,0 +1,72 @@
...
+@pytest.fixture
+def bedrock_client_us_east_1():
+    """Create a BedrockRuntimeClient for us-east-1 (for nova-sonic model)."""
+    return _create_bedrock_client("us-east-1")
+
+
+@pytest.fixture
+def model_id():
+    """Model ID that supports non-streaming and output streaming operations."""
+    return MODEL_ID
```
nateprewitt: Similar note here, if these are constants do they need to be fixtures?

clients/aws-sdk-bedrock-runtime/tests/integration/test_non_streaming.py
jonathan343: Thank you for doing the tests. However, I think it would be sufficient to check that the content in the response is not empty. The idea here is that we're testing that we can make a successful request to the service and receive some non-empty response. It doesn't matter as much what the content of the response is, especially with LLMs which are not deterministic.

clients/aws-sdk-bedrock-runtime/tests/integration/conftest.py
jonathan343: I kinda agree with this. It may also make the tests confusing to read for newer developers who don't know about pytest fixtures. It requires they know about fixtures and that they're defined in `conftest.py`. I remember being confused by them when I first started working on botocore

clients/aws-sdk-bedrock-runtime/tests/integration/__init__.py
```diff
@@ -1,2 +1,26 @@
...
+
+from smithy_aws_core.identity import EnvironmentCredentialsResolver
+
+from aws_sdk_bedrock_runtime.client import BedrockRuntimeClient
+from aws_sdk_bedrock_runtime.config import Config
+
+MODEL_ID = "amazon.titan-text-express-v1"
+BIDIRECTIONAL_MODEL_ID = "amazon.nova-sonic-v1:0"
+MESSAGE = "Who created the Python programming language?"
+AUDIO_FILE = Path(__file__).parent / "fixtures" / "test.pcm"
```
jonathan343: nit - I don't think putting `test.pcm` under the `fixtures` folder makes sense anymore. Maybe something like `assets` makes more sense

jonathan343 (approved):
LGTM! Thanks!

clients/aws-sdk-bedrock-runtime/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,280 @@
...
+    all_text_output: list[str] = []
+
+    await stream.await_output()
+    output_stream = stream.output_stream
+    if output_stream is None:
+        return got_text, got_audio, all_text_output
+
+    async for out in output_stream:
+        if not isinstance(out, InvokeModelWithBidirectionalStreamOutputChunk):
+            continue
```
nateprewitt: What else are we receiving in the stream that we're skipping here?

alexgromero: The stream can yield both normal OutputChunk events and modeled error events (InternalServerException, ValidationException, ThrottlingException, etc.). Those error types don't have a bytes_ payload, so the isinstance check narrows the type and keeps pyright happy.

nateprewitt: Can this be successful then if we get an error chunk as the last part of our stream and just ignore it? I don't immediately remember how termination of the stream works.

alexgromero: Good catch. You're right that if we receive error events after getting some successful chunks the test would still pass since we just continue and the assertions only check that we got some output. Should we fail explicitly on error events? or is it acceptable for the stream to send partial results followed by an error?

nateprewitt: A successful stream shouldn't terminate with errors so I think we'd want to consider that a failure case. I know there have been a lot of unintuitive behaviors with the Bedrock APIs and how they surface errors. I would not be surprised if we get back partial data before it fails, so we should account for that in the test.

If we're seeing typing issues, let's take a look at that. I wonder if we're missing partial definitions or type unions somewhere.

nateprewitt (approved):
Thanks @alexgromero!

PR #38, opened 2025-12-02
Add integration tests for Transcribe Streaming

alexgromero:
*Issue #, if available:*

*Description of changes:*
This PR adds comprehensive integration tests for Amazon Transcribe Streaming client covering the following streaming patterns:
- Non-streaming operation using `get_medical_scribe_stream` (requires setup)
- Bidirectional streaming operation using `start_stream_transcription`

Non-streaming Test Setup:
The `get_medical_scribe_stream` test requires AWS resources (IAM role + S3 bucket). A setup script is included at `tests/setup_resources.py` that creates these resources. The test fails if the required environment variables (`HEALTHSCRIBE_ROLE_ARN` and `HEALTHSCRIBE_S3_BUCKET`) are not set.


clients/aws-sdk-transcribe-streaming/tests/setup_resources.py
```diff
@@ -0,0 +1,93 @@
+# /// script
```
nateprewitt, line 1: This is great, thanks for putting it together! One minor thing is we may want to put this under the `tests/` directory. It may also be worth adding a quick disclaimer that this is for local testing but shouldn't be relied on in production setups.

clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,112 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test bidirectional streaming duplex event stream handling."""
```
nateprewitt: nit; we can probably shorten in this context. While duplex is subtly different than bidirectional, in this context they're representing the same concept. That we can communicate on the stream in both directions.

```suggestion
"""Test bidirectional event stream handling."""
```

clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,112 @@
...
+                    for alt in result.alternatives:
+                        if alt.transcript:
+                            transcripts.append(alt.transcript)
+
+    return got_transcript_events, transcripts
+
+
+async def test_start_stream_transcription() -> None:
+    """Test bidirectional streaming with audio input and transcription output."""
+    transcribe_client = create_transcribe_client("us-west-2")
```
nateprewitt, line 96: We're setting up resources in `us-east-1` but running the client in `us-west-2`. Is there a reason for that or do we have any concern the setup may not work as intended?

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
```diff
@@ -0,0 +1,121 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Test non-streaming output type handling.
```
nateprewitt, line 4: > Test non-streaming output type handling.

I'm not sure I understand what this means 😅 That we're explicitly testing `GetMedicalScribeStreamOutput`? We may be able to consolidate these tests, I don't know if separate files are needed. That will also let us share some of the constants as well.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
```diff
@@ -0,0 +1,121 @@
...
+CHANNEL_NUMS = 1
+CHUNK_SIZE = 1024 * 8
+
+
+async def test_get_medical_scribe_stream() -> None:
+    role_arn = os.environ.get("HEALTHSCRIBE_ROLE_ARN")
+    s3_bucket = os.environ.get("HEALTHSCRIBE_S3_BUCKET")
+
+    if not role_arn or not s3_bucket:
+        pytest.skip("HEALTHSCRIBE_ROLE_ARN or HEALTHSCRIBE_S3_BUCKET not set")
```
nateprewitt: Should this be skipped or fail? It may be easy to miss since this will produce a "successful" run when it wasn't.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
```diff
@@ -0,0 +1,121 @@
...
+                type=MedicalScribeSessionControlEventType.END_OF_SESSION
+            )
+        )
+    )
+    await stream.input_stream.close()
+
+    await stream.await_output()
+    if stream.output_stream:
+        async for _ in stream.output_stream:
+            pass
```
nateprewitt, line 109: This is just to drain the stream so we release the connection?

nateprewitt (changes requested):
Few minor comments but otherwise looks like we're all set. Thanks @alexgromero!

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
alexgromero: Yep!

nateprewitt: Ok cool, this is somewhere we may do a little comment either on the same line or immediately above just saying what we're doing. It's clear when you've seen it before, but may not be to someone newer to the SDK.

clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py
alexgromero: We set up the test resources in `us-east-1` for the medical scribe test (test_non_streaming.py) because Amazon HealthScribe streaming is only available in us-east-1 (it doesn't work in other US regions). However, this specific test doesn't use those resources and I set it to run in `us-west-2` to align with where our infrastructure will be set up.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
alexgromero: Yeah, I meant that we're testing the non-streaming operation which returns a regular response object `GetMedicalScribeStreamOutput`. 

I separated them to match the bedrock runtime [pattern](https://github.com/awslabs/aws-sdk-python/tree/develop/clients/aws-sdk-bedrock-runtime/tests/integration) which has separate files for different streaming types (test_non_streaming.py, test_output_streaming.py, test_bidirectional_streaming.py), but I can combine them if you'd prefer?

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
alexgromero: I chose skip because the test requires external resources that may not be available locally yet. However, I see your point. I can change this to fail by default instead.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
nateprewitt: Yeah, I think I missed that in the bedrock review. Generally, I don't know if they need that degree of separation. I don't feel strongly about it though. We can leave these in-sync with bedrock and revisit later.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
nateprewitt: I'd vote for failure to be explicit, otherwise it gives a false sense of confidence imo. If you, or other team members feel strongly though, we can definitely talk about it further.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
```diff
@@ -47,8 +45,8 @@ async def test_get_medical_scribe_stream() -> None:
     role_arn = os.environ.get("HEALTHSCRIBE_ROLE_ARN")
     s3_bucket = os.environ.get("HEALTHSCRIBE_S3_BUCKET")
 
-    if not role_arn or not s3_bucket:
-        pytest.skip("HEALTHSCRIBE_ROLE_ARN or HEALTHSCRIBE_S3_BUCKET not set")
+    assert role_arn, "HEALTHSCRIBE_ROLE_ARN environment variable not set"
+    assert s3_bucket, "HEALTHSCRIBE_S3_BUCKET environment variable not set"
```
nateprewitt: We can definitely do this approach, what I was asking about earlier was using `pytest.fail` instead of `pytest.skip` though. The explicit test failure let's us surface the configuration issue in one message rather than having to run the script multiple times for the asserts.

The asserts are there to enforce something about the test rather than setup ideally.

nateprewitt (commented):
Last minor comment, otherwise we're ready to merge I think.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
alexgromero: Ahh got it! Updated in latest revision.

nateprewitt (approved):
:shipit:

PR #40, opened 2026-01-02
Release: AWS SDK Bedrock Runtime 0.3.0

nateprewitt:
This PR releases the latest copy of Bedrock Runtime. This release includes **breaking changes** so please review carefully before updating if you're using interceptors. This change will apply to all `0.3.0` releases of clients for `aws-sdk-python`.

### API Changes
* Adds support for Audio Blocks and Streaming Image Output plus new Stop Reasons of malformed_model_output and malformed_tool_use.
* Adds support for Bedrock Runtime Reserved Service.

### Breaking
* Function signature for `resolve_retry_strategy` has been changed to prevent unnecessary code duplication in operation methods. This will affect all 0.3.0 clients.

### Dependencies
* **Updated**: `smithy_aws_core[eventstream, json]` from `~=0.2.0` to `~=0.3.0`.
* **Updated**: `smithy_core` from `~=0.2.0` to `~=0.3.0`.

### Enhancements
* Add comprehensive integration tests for non-streaming, output streaming, and bidirectional streaming operations.


clients/aws-sdk-bedrock-runtime/src/aws_sdk_bedrock_runtime/_private/schemas.py
```diff
@@ -2520,17 +2636,6 @@
     members={"member": {"target": DOCUMENT_CONTENT_BLOCK}},
 )
 
-S3_LOCATION = Schema.collection(
```
nateprewitt, line 2523: I don't immediately understand why this moved. It may be a non-issue but potentially something to check in on for determinism on where shapes go in the file. We may need to enforce ordering.

jonathan343 (approved):

PR #45, opened 2026-01-04
Add changelog management tooling and backfill existing release history

jonathan343:
> [!NOTE]
> These scripts are adapted from [smithy-lang/smithy-python#552](https://github.com/smithy-lang/smithy-python/pull/552), with the addition of support for release summary entries (`SUMMARY.json`).

## Summary

- Add Python scripts for managing per-package changelogs in a structured, automation-friendly format.
- Backfill changelog entries for existing client packages (`aws-sdk-bedrock-runtime`, `aws-sdk-sagemaker-runtime-http2`, `aws-sdk-transcribe-streaming`).

## Changelog Tooling

This PR introduces three scripts under `scripts/changelog/`:

| Script | Purpose |
|--------|---------|
| `new-entry.py` | Create individual changelog entries stored in `.changes/next-release/` |
| `new-release.py` | Consolidate pending entries into a versioned JSON file (e.g., `0.3.0.json`) |
| `render.py` | Generate `CHANGELOG.md` from version JSON files using a Jinja2 template |

### Workflow

1. Contributors add entries via `new-entry.py --package <name> --type <type> --description "..."`
2. At release time, `new-release.py` consolidates entries into `<version>.json`
3. `render.py` regenerates `CHANGELOG.md` from all version files

## Test plan

- [x] Verify `new-entry.py` creates entries in the correct location
- [x] Verify `new-release.py` consolidates entries and cleans up `next-release/`
- [x] Verify `render.py` generates accurate `CHANGELOG.md` output
- [x] Confirm rendered changelogs match the JSON source data

---


clients/aws-sdk-bedrock-runtime/.changes/0.0.1.json
ubaskota, line 1: ~~Are these logs (in `sth.json` files) for the changes that were made in the past?~~ NVM its listed in the summary.

scripts/changelog/new-entry.py
```diff
@@ -0,0 +1,127 @@
...
+    description: str,
+    package_name: str,
+) -> str:
+    # Get package .changes directory and ensure it exists
+    changes_dir = get_package_changes_dir(package_name)
+    changes_dir.mkdir(exist_ok=True)
+
+    # Create next-release directory for pending changes
+    next_release_dir = changes_dir / "next-release"
+    next_release_dir.mkdir(exist_ok=True)
```
ubaskota: Both `create_change_entry` and `create_summary_entry` functions repeat the same logic for directory setup. Do you think they can be consolidated into a single function?

scripts/changelog/new-entry.py
```diff
@@ -0,0 +1,127 @@
...
+    return str(entry_file)
+
+
+def create_summary_entry(
+    description: str,
+    package_name: str,
+) -> str:
+    """Create or update the release summary for the next release."""
+    # Get package .changes directory and ensure it exists
+    changes_dir = get_package_changes_dir(package_name)
```
ubaskota: Will there be a case where the `package_name` is invalid? Do we need to add a validation?

scripts/changelog/render.py
```diff
@@ -0,0 +1,156 @@
...
+    """Get sorted list of version numbers from .changes directory."""
+    version_pattern = re.compile(VERSION_FILE_PATTERN)
+    versions: list[str] = []
+
+    for file in changes_dir.iterdir():
+        if file.is_file() and version_pattern.match(file.name):
+            versions.append(file.stem)
+
+    # Sort by semantic version (oldest first)
+    versions.sort(key=lambda v: [int(x) for x in v.split(".")])
```
ubaskota: Nit: Would using `versions.sort(key=.., reverse=True`) be better than reversing the sorted list inside `render_changes`?

scripts/changelog/new-entry.py
```diff
@@ -0,0 +1,127 @@
...
+    changes_dir.mkdir(exist_ok=True)
+
+    # Create next-release directory for pending changes
+    next_release_dir = changes_dir / "next-release"
+    next_release_dir.mkdir(exist_ok=True)
+
+    # Generate unique filename
+    timestamp = int(time.time() * 1_000_000)
+    unique_id = uuid.uuid4().hex
+    filename = f"{package_name}-{change_type}-{timestamp}-{unique_id}.json"
```
ubaskota: For filename, timestamp is already in microseconds. Wouldn't that be enough? What is the benefit of adding uuid to it?

ubaskota (commented):
Overall looks good. I've added some comments. Please, let me know if you have any questions.

scripts/changelog/new-entry.py
jonathan343: Relying on timestamps alone will never really guarantee uniqueness since our infra will be creating changelog entries in rapid succession (and in parallel in the future). See the example below:

```python
>>> timestamps = []
>>> for i in range(1000):
...     timestamp = int(time.time() * 1_000_000)
...     if timestamp in timestamps:
...         print("Found duplicate")
...     else:
...         timestamps.append(timestamp)
...         
Found duplicate
Found duplicate
Found duplicate
```

We could increase the precision, however, using uuid is more robust. I decided to include the timestamp in attempt to achieve natural ordering of changelog entries when rendering them out. However, that isn't really working right now and I'll handle it in a separate PR. It's more of a nice to have. 

I'm gonna remove timestamp for now and only use uuid.

scripts/changelog/render.py
jonathan343: Either one works tbh. I'll add a `reverse` parameter to `get_sorted_versions` and provide `True` when we call it

scripts/changelog/new-entry.py
jonathan343: If it doesn't exist, we'll get the following error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/gytndd/dev/GitHub/aws-sdk-python/clients/aws-sdk-bedrock-runtimes/.changes'
```

I can add a better error message

clients/aws-sdk-bedrock-runtime/.changes/0.3.0.json
```diff
@@ -0,0 +1,28 @@
...
+      "type": "breaking",
+      "description": "Function signature for `resolve_retry_strategy` has been changed to prevent unnecessary code duplication in operation methods. This will affect all 0.3.0 clients."
+    },
+    {
+      "type": "enhancement",
+      "description": "Add comprehensive integration tests for non-streaming, output streaming, and bidirectional streaming operations."
+    },
+    {
+      "type": "dependency",
+      "description": "**Updated**: `smithy_aws_core[eventstream, json]` from `~=0.2.0` to `~=0.3.0`."
```
SamRemis, line 21: This seems like an odd thing to bold to begin with, and it is less readable (for me) in both JSON and markdown.

Not a huge thing, and we shouldn't change them retroactively, but if we are going to start using markdown syntax in changelog entries, we should agree on guidelines for when and how to apply them and record them in our contributing.md file.

scripts/changelog/new-entry.py
```diff
@@ -0,0 +1,127 @@
...
+import uuid
+from pathlib import Path
+
+PROJECT_ROOT_DIR = Path(__file__).resolve().parent.parent.parent
+
+
+def get_package_changes_dir(package_name: str) -> Path:
+    package_path = PROJECT_ROOT_DIR / "clients" / package_name
+    changes_dir = package_path / ".changes"
+    return changes_dir
```
SamRemis: Nit: 
Personally I don't find this any more readable than a single line function of `return PROJECT_ROOT_DIR / "clients" / package_name /  ".changes"`. 

If you do or prefer this for any reason, keep it as is.

scripts/changelog/new-entry.py
SamRemis: +1; I'm not sure I understand the reason for duplicating this logic.  It's minor, but this could easily be one function:

```
def ensure_changes_dir(package_name):
    changes_dir = PROJECT_ROOT_DIR / "clients" / package_name / ".changes"
    changes_dir.mkdir(parents=True, exist_ok=True)
    return changes_dir
```

scripts/changelog/new-entry.py
```diff
@@ -0,0 +1,127 @@
...
+    print(f"Created release summary: {summary_file}")
+    return str(summary_file)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Create a new changelog entry")
+    parser.add_argument(
+        "-t",
+        "--type",
+        # TODO: Remove the 'breaking' option once this project is stable.
```
SamRemis: Just a thought- maybe we leave this in permanently?  For minor "breaks" that we will continue to do that offer more benefit than pain to customers.

One example that may be upcoming in boto is switching to using `standard` instead of `legacy` retry mode.

scripts/changelog/new-release.py
```diff
@@ -0,0 +1,207 @@
...
+        print(f"Error: Version file {version_file} already exists!")
+        sys.exit(1)
+
+    version_data: dict[str, Any] = {"changes": changes}
+
+    if summary:
+        version_data["summary"] = summary
+
+    with open(version_file, "w") as f:
+        json.dump(version_data, f, indent=2)
```
SamRemis, line 117:
```suggestion
        json.dump(version_data, f, indent=2)
        f.write("\n")
```

Adding a newline at the end of the file will prevent git from showing the warning and help with tools that prefer files end in newlines.

scripts/changelog/new-release.py
```diff
@@ -0,0 +1,207 @@
...
+def cleanup_next_release_dir(next_release_dir: Path) -> int:
+    removed_count = 0
+
+    for entry_file in next_release_dir.iterdir():
+        if entry_file.is_file() and entry_file.suffix == ".json":
+            try:
+                entry_file.unlink()
+                removed_count += 1
+            except OSError as e:
+                print(f"Warning: Could not remove {entry_file}: {e}", file=sys.stderr)
```
SamRemis: The intention here is to ignore the error quietly and continue?  Do you think it's possible that this error will get ignored and leave behind the entry file and duplicate the entry for the next release?

clients/aws-sdk-bedrock-runtime/.changes/0.3.0.json
jonathan343: Yea, we can get something added in CONTRIBUTING.md. That should probably be done in a separate PR since there are non-changelog things we should also update it with.

I'm curious what about the bold syntax makes it less readable for you?

scripts/changelog/new-entry.py
jonathan343: Updated this to print the following error message:

```
Error: Package 'foo-bar' not found in clients directory
```

scripts/changelog/new-entry.py
jonathan343: I'm not against it. I think there should be some confirmation that explains the effects of using `breaking` and that it should be used sparingly. Probably wouldn't hurt to add the same for a `feature` since they both result in minor bumps across the ecosystem.

I'll update to:
```py
# TODO: Prompt the user for confirmation before allowing the 'breaking' or `feature` options.
```

scripts/changelog/new-release.py
jonathan343: I think this would be a very rare case when this happens (if possible) because of how we're using this. I'll remove the try/except so if this does raise an error, our script and release process fail

scripts/changelog/new-entry.py
jonathan343: Added the shared setup to a new `setup_changes_directories` function.

clients/aws-sdk-bedrock-runtime/.changes/0.3.0.json
SamRemis: Especially in JSON, it highlights it awkwardly for me  It's kind of like how it's a little harder to read all caps.  IMO each of these updates should be highlighted equally.

SamRemis (approved):

PR #51, opened 2026-04-03
Add generated Lex Runtime V2 client with integration tests

Alan4506:
*Description of changes:*

This PR adds support for the Lex Runtime V2 service by adding the latest Smithy model and code-generating a new client package (`aws-sdk-lex-runtime-v2`). It also adds integration tests covering the following patterns:

- Non-streaming operation using `recognize_text`
- Bidirectional streaming operation using `start_conversation` (TEXT mode)

A setup script at `tests/setup_resources.py` creates the required AWS resources (IAM role, Lex V2 bot, en_US locale with a Greeting intent and closing response).

This branch will be rebased once the release automation publishes the client. 


clients/aws-sdk-lex-runtime-v2/.changes/next-release/aws-sdk-lex-runtime-v2-api-change-fa23e4248cb149cfacb71f203f455e83.json
```diff
@@ -0,0 +1,4 @@
+{
```
arandito: Do we need this changelog entry? Our release automation should handle new client setup, including initial versioning and changelog entries. We will have merge conflicts if the wording if different or we may have repeated entries because the uuid in the newly generated entry is different.

clients/aws-sdk-lex-runtime-v2/CHANGELOG.md
arandito, line 1: Same comment as before. This should be done by our release automation.

clients/aws-sdk-lex-runtime-v2/README.md
```diff
@@ -0,0 +1,10 @@
+# Code generated by smithy-python-codegen DO NOT EDIT.
```
arandito: Anyway we can avoid this header on the package README? This will also appear on the PyPI page for the package.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        _wait_for_bot_locale(lex, bot_id, locale_id, target_status="Built")
+
+    # Use TSTALIASID (test alias, always available)
+    bot_alias_id = "TSTALIASID"
+
+    return bot_id, bot_alias_id, locale_id
+
+
+def _wait_for_bot(lex, bot_id: str, timeout: int = 60) -> None:
```
arandito: We should use our waiters instead of a custom wait function. Waiters poll AWS resources until they become available so it's meant for cases like these. According to the [boto3 docs](https://docs.aws.amazon.com/boto3/latest/reference/services/lexv2-models/waiter/BotAvailable.html), we have a `bot_available` waiter that waits until a bot is available after its created. We can probably do something like this:

```python
response = lex.create_bot(
    botName=bot_name,
    roleArn=role_arn,
    dataPrivacy={"childDirected": False},
    idleSessionTTLInSeconds=300,
)
bot_id = response["botId"]
print(f"Created bot: {bot_id}")

lex.get_waiter("bot_available").wait(
    botId=bot_id,
    WaiterConfig=WAITER_CONFIG,
)
```

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+        status = response["botStatus"]
+        if status == "Available":
+            return
+        if status in ("Failed", "Deleting"):
+            raise RuntimeError(f"Bot creation failed with status: {status}")
+        time.sleep(5)
+    raise TimeoutError("Bot did not become available")
+
+
+def _wait_for_bot_locale(
```
arandito: Same here, we should be should be using waiters.

I see this functions handles two cases: 1) checking that the locale is created  and 2) checking that it's built. These two cases should be handled by the [`bot_locale_created`](https://docs.aws.amazon.com/boto3/latest/reference/services/lexv2-models/waiter/BotLocaleCreated.html) and [`bot_locale_built`](https://docs.aws.amazon.com/boto3/latest/reference/services/lexv2-models/waiter/BotLocaleBuilt.html) waiters.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        locale_status = locale_resp["botLocaleStatus"]
+        print(f"Locale status: {locale_status}")
+    except lex.exceptions.ResourceNotFoundException:
+        print("Creating locale...")
+        lex.create_bot_locale(
+            botId=bot_id,
+            botVersion="DRAFT",
+            localeId=locale_id,
+            nluIntentConfidenceThreshold=0.40,
```
arandito: Not too familiar with the Lex v2 API. Why did we choose 0.40? We should have a comment explaining it or make it a constant.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    if existing_bots["botSummaries"]:
+        bot_id = existing_bots["botSummaries"][0]["botId"]
+        print(f"Bot already exists: {bot_id}")
+    else:
+        response = lex.create_bot(
+            botName=bot_name,
+            roleArn=role_arn,
+            dataPrivacy={"childDirected": False},
+            idleSessionTTLInSeconds=300,
```
arandito: Any specific reason we chose 300 seconds?

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+        if status == target_status:
+            return
+        if status in ("Failed", "Deleting"):
+            raise RuntimeError(f"Bot locale failed with status: {status}")
+        time.sleep(5)
+    raise TimeoutError(f"Bot locale did not reach {target_status}")
+
+
+if __name__ == "__main__":
+    bot_id, bot_alias_id, locale_id = create_lex_bot()
```
arandito: Why do we return `bot_id`, `bot_alias_id`, and `locale_id` if we only print the `bot_id` for the tests?

clients/aws-sdk-lex-runtime-v2/tests/integration/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+import os
+
+from smithy_aws_core.identity import EnvironmentCredentialsResolver
+
+from aws_sdk_lex_runtime_v2.client import LexRuntimeV2Client
+from aws_sdk_lex_runtime_v2.config import Config
+
+BOT_ID = os.environ["LEX_BOT_ID"]
+BOT_ALIAS_ID = "TSTALIASID"
+LOCALE_ID = "en_US"
```
arandito, line 10: nit: We are setting these up again instead of using the output from `setup_resources.py` which can introduce some drift.

clients/aws-sdk-lex-runtime-v2/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14"
+]
+
+dependencies = [
+    "smithy_aws_core[eventstream, json]~=0.4.0",
+    "smithy_core~=0.3.0",
```
arandito: When I pulled down these changes, installed the client locally, and tried to run the integ tests, I got this error:
```
TypeError: APIOperation.__init__() got an unexpected keyword argument 'error_schemas'
```

This is because the `error_schemas` field was introduced in [`smithy-core 0.4.0`](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-core/CHANGELOG.md#v040) which was released on Tuesday. However, the codegen changes to generate these were merged as part of this [commit](https://github.com/smithy-lang/smithy-python/commit/6ac602204eab4feff05ad87b6b5fc67b8e66f43a) on 3/31.

I think this client was created in between the actual release of smithy-core and the codegen merge. Let's regenerate the client using the latest release to make sure all dependencies are up-to-date for testing.

clients/aws-sdk-lex-runtime-v2/tests/integration/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+from smithy_aws_core.identity import EnvironmentCredentialsResolver
+
+from aws_sdk_lex_runtime_v2.client import LexRuntimeV2Client
+from aws_sdk_lex_runtime_v2.config import Config
+
+BOT_ID = os.environ["LEX_BOT_ID"]
```
arandito: I prefer if we retrieved the env variables in the tests themselves with error handling for cases where the user did not set up the resources.

Right now if I run these tests without setup I get:
```
================================== ERRORS ==================================
____ ERROR collecting tests/integration/test_bidirectional_streaming.py ____
../../../../../.local/share/uv/python/cpython-3.12.9-macos-aarch64-none/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1310: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:488: in _call_with_frames_removed
    ???
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
<frozen importlib._bootstrap_external>:999: in exec_module
    ???
<frozen importlib._bootstrap>:488: in _call_with_frames_removed
    ???
tests/integration/__init__.py:11: in <module>
    BOT_ID = os.environ["LEX_BOT_ID"]
             ^^^^^^^^^^^^^^^^^^^^^^^^
<frozen os>:714: in __getitem__
    ???
E   KeyError: 'LEX_BOT_ID'
========================= short test summary info ==========================
ERROR tests/integration/test_bidirectional_streaming.py - KeyError: 'LEX_BOT_ID'
!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!
============================= 1 error in 0.15s =============================
```
The error is not handled cleanly and the user has to look for where `LEX_BOT_ID` is set, which isn't in the test.

If I gracefully handle it in the test similar to our [transcribe streaming tests](https://github.com/awslabs/aws-sdk-python/blob/develop/clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py#L47-L51):
```
================================= FAILURES =================================
_________________________ test_start_conversation __________________________

    async def test_start_conversation() -> None:
        """Test bidirectional streaming StartConversation operation."""
        client = create_lex_client("us-east-1")
        bot_id = os.environ.get("LEX_BOT_ID")
        if not bot_id:
>           pytest.fail("LEX_BOT_ID environment variable not set")
E           Failed: LEX_BOT_ID environment variable not set

tests/integration/test_bidirectional_streaming.py:98: Failed
========================= short test summary info ==========================
FAILED tests/integration/test_bidirectional_streaming.py::test_start_conversation - Failed: LEX_BOT_ID environment variable not set
============================ 1 failed in 0.13s =============================
```
The failure point is clear and easier to debug.

clients/aws-sdk-lex-runtime-v2/tests/integration/test_non_streaming.py
```diff
@@ -0,0 +1,39 @@
...
+    assert response.session_state.intent is not None
+    assert response.session_state.intent.name == "Greeting"
+
+    # Verify interpretations contain the matched intent
+    assert response.interpretations is not None
+    assert len(response.interpretations) > 0
+
+    intent_names = [i.intent.name for i in response.interpretations if i.intent]
+    assert "Greeting" in intent_names
+    assert "FallbackIntent" in intent_names
```
arandito: Is this fallback intent always returned?

clients/aws-sdk-lex-runtime-v2/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    got_text_response = False
+    messages: list[str] = []
+
+    _, output_stream = await stream.await_output()
+    if output_stream is None:
+        return got_text_response, messages
+
+    async for event in output_stream:
+        if isinstance(event, StartConversationResponseEventStreamTextResponseEvent):
```
arandito: Why do we not raise a runtime error when there is an unexpected event similar to current bidi integ tests: https://github.com/awslabs/aws-sdk-python/blob/develop/clients/aws-sdk-bedrock-runtime/tests/integration/test_bidirectional_streaming.py#L221-L223

Alan4506: Your are right to point this out. And I'll add checks for other possible events as well.

clients/aws-sdk-lex-runtime-v2/tests/integration/__init__.py
SamRemis: Thanks for the great write-up here, Antonio :) 

While it has it's merits, personally I don't love the idea of having environment variables required for our integ tests at all.  If we start adding a bunch of these, they can get out of hand really quickly.  It's easy to imagine a scenario where we have 
`LEX_BOT_ID`, but also `LEX_BOT_S3_OUTPUT_BUCKET`, which is different from `TRANSCRIBE_STREAMING_OUTPUT_S3_BUCKET`, and just have a mounting list of resources that we all have to make and set in our environment before running the test.

These tests should be self-contained whenever possible.  Have we discussed before what our philosophy is for when we should do each of the following?
1. Add a new resource via environment variable (as is done in this PR)
2. Skip an integ test entirely
3. Mock service responses instead of relying on AWS resources and make it a functional test instead of an integ test

IMO there's a tradeoff here, and I'd personally rather see these as functional as opposed to integ tests.  This both saves money and improves security for anyone running these tests.

clients/aws-sdk-lex-runtime-v2/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,108 @@
...
+            locale_id=LOCALE_ID,
+            session_id=str(uuid.uuid4()),
+            conversation_mode="TEXT",
+        )
+    )
+
+    results = await asyncio.gather(_send_events(stream), _receive_events(stream))
+
+    got_text_response, messages = results[1]
+    assert got_text_response, "Expected to receive a TextResponse event"
```
SamRemis: Nit: I think this assertion is redundant.  got_text_response will never be `True` when `len(messages) > 0` is `False.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
```diff
@@ -0,0 +1,182 @@
...
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Setup script to create AWS resources for Lex Runtime V2 integration tests.
+
+Creates a simple Lex V2 bot with a Greeting intent for testing.
+
+Note:
+    This script is intended for local testing only and should not be used for
```
SamRemis: [Discussion]
This setup script approach is understandable (and we already have a similar pattern elsewhere in the repo), but I’d like to clarify the intended long-term workflow for these integration tests:

1. What’s the recommended teardown/cleanup path after running the tests (locally and/or in CI)? Should the script also provide a destroy mode?
2. For recurring runs (e.g., monthly), do we expect resources to be recreated each time, or reused? If reused, where/how should resource identifiers be stored (config file/SSM/etc.)?
3. The doc says “not for production setups” — what environment is expected to run these tests long-term (dedicated test account, ephemeral CI account, etc.)?

If these tests are intended to be temporary, that’s fine — but if they’re long-lived, I think we should document the lifecycle story

clients/aws-sdk-lex-runtime-v2/tests/test_protocol.py
```diff
@@ -0,0 +1,64 @@
...
+        self._client_config = client_config
+
+    async def send(
+        self,
+        request: HTTPRequest,
+        *,
+        request_config: HTTPRequestConfiguration | None = None,
+    ) -> HTTPResponse:
+        # Raise the exception with the request object to bypass actual request handling
+        raise TestHttpServiceError(request)
```
SamRemis, line 34: [Question]

I'm trying to wrap my head around this conceptually - since we have one of these per service, when are we going to move this to a centralized area like smithy-aws-core.  That doesn't need to be done in this PR, but all of this duplication is probably something we should start to consolidate

SamRemis (commented):
Left a few comments - Antonio was very thorough and got most of them.  

To make sure I understand - are we expecting this to be the workflow:
1. A different PR adds the changelog entry and the models
2. We release the client via internal infrastructure
3. After confirming that this PR contains only the changes for the integ tests, we merge this PR so these get run for future releases automatically

Is that all correct?

Alan4506:
> Left a few comments - Antonio was very thorough and got most of them.
> 
> To make sure I understand - are we expecting this to be the workflow:
> 
> 1. A different PR adds the changelog entry and the models
> 2. We release the client via internal infrastructure
> 3. After confirming that this PR contains only the changes for the integ tests, we merge this PR so these get run for future releases automatically
> 
> Is that all correct?

1. A different PR will be made to add the model, but as [the comment](https://github.com/awslabs/aws-sdk-python/pull/51#discussion_r3059308482) says, the changelog entry might be handled by the release automation, so we won't add changelog entry in that different PR.
2. Yes, we will release the client via internal infrastructure
3. The PR may not only contain only the changes for the integ tests. After the client is released, in @jonathan343's opinion, there shouldn't be much difference between the this PR and upstream, except the integration tests, so the merge should be fine. But worst case if the merge conflict does happen, we can update this PR to only have the integ tests, or just create a new PR to only have the integ tests. But for now I'll just regenerate the client using the latest release to make sure all dependencies are up-to-date for testing, so that it is easier for reviewers to review the tests.

clients/aws-sdk-lex-runtime-v2/tests/test_protocol.py
Alan4506: It makes sense to me. In the future we should consider how to handle this duplication.

clients/aws-sdk-lex-runtime-v2/.changes/next-release/aws-sdk-lex-runtime-v2-api-change-fa23e4248cb149cfacb71f203f455e83.json
Alan4506: Removed.

clients/aws-sdk-lex-runtime-v2/CHANGELOG.md
Alan4506: Removed.

clients/aws-sdk-lex-runtime-v2/README.md
Alan4506: We need to add LICENSE and NOTICE files in Catapult for new packages so maybe we can do the same for README. So in the future I may delete the README file from this PR, but for now I'll keep it because `uv pip install -e .` will cause en error if the README does not exist.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
Alan4506: Great suggestion. I've used waiters in my new revision.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
Alan4506: Same as above. I've added waiters in my new revision.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
Alan4506: Comment added to explain it.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
Alan4506: Comment added to explain it.

clients/aws-sdk-lex-runtime-v2/tests/integration/__init__.py
Alan4506: `setup_resources.py` has been removed in my new revision, and these values are used by multiple files, so I think it makes sense to keep them here?

clients/aws-sdk-lex-runtime-v2/pyproject.toml
Alan4506: I've regenerated the client using the latest release.

clients/aws-sdk-lex-runtime-v2/tests/integration/__init__.py
Alan4506: `setup_resources.py` has been removed. Now resource setup and teardown is handled by a session-scoped pytest fixture in conftest.py, which creates and deletes the Lex bot automatically when running integration tests. We no longer need environment variables.

clients/aws-sdk-lex-runtime-v2/tests/integration/test_non_streaming.py
Alan4506: Yes, their documentation suggests this. And I also verified it by running the test hundreds of times - fallback intent always returned for our test case.

clients/aws-sdk-lex-runtime-v2/tests/integration/test_bidirectional_streaming.py
Alan4506: I've made a new revision for this file. Now we validate if all 4 expected event types (TranscriptEvent, IntentResultEvent, TextResponseEvent, HeartbeatEvent) can be received.

clients/aws-sdk-lex-runtime-v2/tests/setup_resources.py
Alan4506: Please see my reply to another comment:
https://github.com/awslabs/aws-sdk-python/pull/51#discussion_r3089653328

clients/aws-sdk-lex-runtime-v2/tests/integration/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+                "Action": "sts:AssumeRole",
+            }
+        ],
+    }
+    try:
+        iam.create_role(
+            RoleName=ROLE_NAME, AssumeRolePolicyDocument=json.dumps(trust_policy)
+        )
+    except iam.exceptions.EntityAlreadyExistsException:
+        pass
```
SamRemis: [Change requested]
There's a race condition for running these in parallel in this account:

1. Runner 1 creates the role and succeeds
2. Runner 2 tries to create it, but reaches this pass statement
3. Runner 1 starts the teardown and deletes the iam role
4. Runner 2 fails the tests since the IAM role does not exist.

We may want to make the name randomly generated here?  Or if there's an existing AWS-owned role we can use, that would be great too.

SamRemis (changes requested):

clients/aws-sdk-lex-runtime-v2/tests/integration/conftest.py
Alan4506: Good catch! According to the [CreateBot API docs](https://docs.aws.amazon.com/lexv2/latest/APIReference/API_CreateBot.html), `roleArn` is a required parameter and I cannot find any pre-existing AWS-managed role we can use for this.

The same race condition also applies to bot names: the docs mention ["The bot name must be unique in the account that creates the bot"](https://docs.aws.amazon.com/lexv2/latest/APIReference/API_CreateBot.html).

Fixed both by appending a unique UUID suffix to the role name and bot name.

clients/aws-sdk-lex-runtime-v2/tests/integration/conftest.py
```diff
@@ -0,0 +1,158 @@
...
+from typing import Any
+
+import boto3
+import pytest
+
+from . import LOCALE_ID, REGION
+
+# Tags applied to all resources so orphaned resources from interrupted
+# test runs can be discovered and cleaned up.
+_TAGS = [{"Key": "Purpose", "Value": "IntegTest"}]
```
SamRemis, line 22: [Non-blocking]

Since the idea behind this is to have a single tag that can be applied to as many integ test related resources as possible, this would be better suited for `smithy-aws-core` and we can import it.  

Since that would require a PR to smithy-aws-core and a publish, can we just make a backlog item for that and come back?

SamRemis (approved):

PR #54, opened 2026-04-23
Add generated Q Business client and integration tests

Alan4506:
*Description of changes:*

This PR adds integration tests for the Q Business service client (`aws-sdk-qbusiness`), covering the following:

- Non-streaming operations: `create_application`, `get_application`, `delete_application`, `create_index`, `get_index`, `create_retriever`, `get_retriever`,  `chat_sync`.
- Bidirectional event streaming operation: `chat`.

The test structure and patterns follow the same conventions established in the Lex Runtime V2 integration tests (PR #51), such as session-scoped conftest fixtures for resource setup/teardown.

To successfully run the tests locally, follow the steps below:
```
# Set up your AWS credentials, then:
cd <path-to-smithy-python>
git checkout develop && git pull
make build-py

# Set up and run Q Business integration tests
cd <path-to-aws-sdk-python>/clients/aws-sdk-qbusiness
uv venv .venv && source .venv/bin/activate
uv pip install --group test -e . 
python -m pytest tests/integration
```

All tests should pass.


clients/aws-sdk-qbusiness/tests/integration/conftest.py
```diff
@@ -0,0 +1,156 @@
...
+    qbusiness = boto3.client("qbusiness", region_name=REGION)
+    qbusiness.delete_application(applicationId=app_id)
+
+
+@pytest.fixture(scope="session")
+def qbusiness_app():
+    """Create a Q Business application for the test session and delete it after."""
+    app_id = _create_qbusiness_app()
+    yield app_id
+    _delete_qbusiness_app(app_id)
```
ubaskota: If `_create_qbusiness_app()` fails after the application is created (during index or retriever creation), will `_delete_qbusiness_app` still run to clean up the application?

clients/aws-sdk-qbusiness/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,111 @@
...
+
+    await stream.input_stream.send(
+        ChatInputStreamTextEvent(value=TextInputEvent(user_message="Hello"))
+    )
+
+    await stream.input_stream.send(
+        ChatInputStreamEndOfInputEvent(value=EndOfInputEvent())
+    )
+
+    await asyncio.sleep(1)
```
ubaskota: Why is the 1-second sleep needed after sending the input stream? Could you add a comment explaining the reason?

clients/aws-sdk-qbusiness/tests/integration/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from aws_sdk_qbusiness.config import Config
+
+REGION = "us-east-1"
+
+
+def create_qbusiness_client(region: str) -> QBusinessClient:
+    """Helper to create a QBusinessClient for a given region."""
+    return QBusinessClient(
+        config=Config(
+            endpoint_uri=f"https://qbusiness.{region}.api.aws",
```
ubaskota, line 16: This is for my own understanding:
- Why is the `endpoint_uri` hardcoded here? Is it not possible to resolve it automatically?

Alan4506: Currently we only have a `StandardRegionalEndpointsResolver` in `smithy-aws-core` that resolves `{prefix}.{region}.amazonaws.com`, but Q Business uses `.api.aws` instead of `.amazonaws.com`, so it doesn't work here. See [smithy_aws_core/endpoints/standard_regional.py](https://github.com/smithy-lang/smithy-python/blob/develop/packages/smithy-aws-core/src/smithy_aws_core/endpoints/standard_regional.py#L25-L27).

clients/aws-sdk-qbusiness/tests/integration/test_bidirectional_streaming.py
Alan4506: ~~Added a comment explaining the sleep. It gives the service time to process input events before we close the stream. The same pattern is used in the transcribe-streaming and bedrock-runtime bidirectional streaming tests ([example](https://github.com/awslabs/aws-sdk-python/blob/44dc092d89706412124bf04d33f9378afd81d76f/clients/aws-sdk-transcribe-streaming/tests/integration/test_bidirectional_streaming.py#L56)). Tests will fail if we remove this.~~
Removed. Verified that tests still pass.

clients/aws-sdk-qbusiness/tests/integration/conftest.py
Alan4506: Good catch. Updated the fixture to use `try/finally` so cleanup runs even if creation fails partway through. `_delete_qbusiness_app` now accepts None and returns early if no app was created. Since deleting the application cascades to its index and retriever, this handles all partial failure cases.

clients/aws-sdk-qbusiness/tests/integration/conftest.py
```diff
@@ -0,0 +1,218 @@
...
+            input=GetApplicationInput(application_id=application_id)
+        )
+        if response.status == ApplicationStatus.ACTIVE:
+            return
+        if response.status in {ApplicationStatus.FAILED, ApplicationStatus.DELETING}:
+            raise RuntimeError(
+                f"Application {application_id} entered terminal state {response.status}"
+            )
+        await asyncio.sleep(_POLL_INTERVAL_SECONDS)
+    raise TimeoutError(f"Application {application_id} did not become ACTIVE in time")
```
jonathan343, line 68: Do you know how long this generally takes? If it's well under this timeout, that's fine. If it's can take up to 5min, we may want to be more lenient so we don't leak resources often.

clients/aws-sdk-qbusiness/tests/integration/conftest.py
```diff
@@ -0,0 +1,218 @@
...
+
+async def _wait_for_application_active(client, application_id: str) -> None:
+    """Wait for an Application to reach ACTIVE status.
+
+    Args:
+        client: A Q Business client.
+        application_id: The Application ID to poll.
+    """
+    deadline = asyncio.get_event_loop().time() + _POLL_TIMEOUT_SECONDS
+    while asyncio.get_event_loop().time() < deadline:
```
jonathan343: nit - Looks like `asyncio.get_event_loop()` will be deprecated in the future. Seems like we should use `asyncio.get_running_loop()` instead:

> Note The `asyncio` policy system is deprecated and will be removed in Python 3.16; from there on, this function will return the current running event loop if present else it will return the loop set by [set_event_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.set_event_loop).

clients/aws-sdk-qbusiness/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,110 @@
...
+
+    _, output_stream = await stream.await_output()
+    if output_stream is None:
+        return got_text_events, got_metadata_event
+
+    async for event in output_stream:
+        if isinstance(event, ChatOutputStreamTextEvent):
+            got_text_events = True
+            assert event.value.system_message_type is not None
+            assert event.value.system_message_type == "RESPONSE"
```
jonathan343: nit - Asserting something `is not None` followed by asserting equality seems redundant. Was there a specific reason for this?

jonathan343:
Thanks @Alan4506!

Left some very minor comments above. All tests succeed for me locally.
I'm fine approving once the commends are addressed.

clients/aws-sdk-qbusiness/tests/integration/conftest.py
jonathan343: Forgot to delete this after we discussed

clients/aws-sdk-qbusiness/tests/integration/conftest.py
Alan4506: Thanks! Addressed in https://github.com/awslabs/aws-sdk-python/pull/54/commits/20260484e8ceb037688c743e2850e873b3923c37.

clients/aws-sdk-qbusiness/tests/integration/test_bidirectional_streaming.py
Alan4506: Thanks! Addressed in https://github.com/awslabs/aws-sdk-python/pull/54/commits/20260484e8ceb037688c743e2850e873b3923c37.

jonathan343 (approved):
LGTM! Thanks!

PR #55, opened 2026-04-23
Add conftest fixture to transcribe-streaming integration tests

Alan4506:
*Description of changes:*

This PR refactors Transcribe Streaming integration tests, by replacing the standalone `setup_resources.py` script with a session-scoped pytest fixture in `conftest.py`. This aligns with the pattern of [Lex Runtime V2 integration tests](https://github.com/awslabs/aws-sdk-python/pull/51/changes#diff-b0b1ecd16b4c4233019c6e689c91615594d7631763b05f093a7891c9b50bb65e). The HealthScribe test (`test_get_medical_scribe_stream`) previously required manually running the setup script and exporting environment variables before running tests. Now the IAM role and S3 bucket are automatically created and torn down per test session.

Changes:
- Added `conftest.py` with `healthscribe_resources` fixture (creates IAM role + S3 bucket, tears down after)
- Updated `test_non_streaming.py` to use the fixture instead of environment variables
- Added UUID suffix to role and bucket names to avoid race conditions in parallel runs

To run the integration tests, you should do the following under `clients/aws-sdk-transcribe-streaming`:
```
uv venv .venv
source .venv/bin/activate
uv pip install --group test -e . boto3
python -m pytest tests/integration
```
All tests should pass.


clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+integration tests once per test session. The ``healthscribe_resources``
+fixture provides the role ARN and bucket name.
+"""
+
+import json
+import time
+import uuid
+from typing import Any
+
+import boto3
```
alexgromero, line 15: Are we documenting anywhere that users must `pip install boto3` to run these tests? If customers don't know, they'll just run `uv sync --group test` and get an error like:

```
ImportError while loading conftest '.../aws-sdk-python/clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py'.
tests/integration/conftest.py:16: in <module>
    import boto3
E   ModuleNotFoundError: No module named 'boto3'
```

The previous `setup_resources.py` had inline [PEP 723](https://peps.python.org/pep-0723/) metadata that automatically installed boto3 when running `uv run setup_resources.py`. Now that we've moved to conftest fixtures, boto3 needs to be installed but it is not included in the `pyproject.toml`.

Do you have any thoughts on the best way to handle this? Either documenting it clearly or updating the code generator to include boto3 in test dependencies? For the latter, I'm not sure smithy-python codegen supports per-service test dependencies right now.

Alan4506: I did consider where we should put the logic of adding boto3 dependency. I think we should not add it to our smithy-python codegen since boto3 is only for integration tests that are hand written. In our next release, we will add boto3 as a test dependency in `pyproject.toml` of those clients that need boto3. I've done a release on my dev account to verify that all tests will pass (including the refactored tests in this PR) and boto3 will be added to `pyproject.toml` of transcribe streaming.

clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py
```diff
@@ -0,0 +1,119 @@
...
+    s3 = boto3.client("s3", region_name=REGION)
+    sts = boto3.client("sts")
+
+    account_id = sts.get_caller_identity()["Account"]
+
+    s3.create_bucket(Bucket=BUCKET_NAME)
+    _create_iam_role(iam, ROLE_NAME, BUCKET_NAME)
+
+    # Wait for IAM role to propagate across services.
+    time.sleep(10)
```
alexgromero: I'm a little iffy on the `time.sleep(10)`. IAM is eventually consistent and the [docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/troubleshoot.html#troubleshoot_general_eventual-consistency) explicitly recommend verifying propagation before depending on it, so I feel like a fixed sleep may end up flaking in CI even if it's fine locally.
      
Looking at other SDKs like [botocore](https://github.com/boto/botocore/blob/develop/tests/integration/test_credentials.py#L235-L263) and [Kotlin](https://github.com/aws/aws-sdk-kotlin/blob/fb5222617e42b18c9801f63c04041c45040e370c/tests/benchmarks/service-benchmarks/jvm/src/aws/sdk/kotlin/benchmarks/service/definitions/StsBenchmark.kt#L56), they handle this by retrying the call until the role is actually usable instead of guessing a wait time.

I think something roughly like this inside `test_get_medical_scribe_stream` might be more robust than the fixed sleep.
```
async with asyncio.timeout(300):
    while True:
        try:
            await _run_medical_scribe_session(...)
            return
        except BadRequestException:
            await asyncio.sleep(5)
```

clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py
alexgromero, line 1: A few small things in this file that could be tightened together if you want:
- `boto3.resource` is feature frozen ([ref](https://docs.aws.amazon.com/boto3/latest/guide/resources.html)). The client equivalent (`list_objects_v2` paginator + `delete_objects` + `delete_bucket`) might be a better alternative as it is on the supported path and just a few extra lines.
- The IAM/S3 clients are constructed once in `_create_*` and again in `_delete_*`. We could possibly lift them to the fixture and pass them in.
- `_UNIQUE_SUFFIX`, `ROLE_NAME`, and `BUCKET_NAME` being defined at module scope means the UUID is generated at import time. Moving them into the fixture body would keep the resource lifecycle self-contained to the test session.

clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
```diff
@@ -121,3 +111,23 @@ async def test_get_medical_scribe_stream() -> None:
...
+    """
+    role_arn, s3_bucket = healthscribe_resources
+
+    async with asyncio.timeout(ROLE_PROPAGATION_TIMEOUT):
+        while True:
+            try:
+                await _run_medical_scribe_session(role_arn, s3_bucket)
+                return
+            except BadRequestException:
+                await asyncio.sleep(ROLE_PROPAGATION_RETRY_DELAY)
```
alexgromero: One small thing: when this times out the test fails with a bare `TimeoutError` and the underlying `BadRequestException` won't be in the failure summary. If this ever fails, having the actual error in the traceback makes it much easier to debug. Could we capture the last exception and re-raise it on timeout?


```suggestion
    last_error: BadRequestException | None = None
    try:
        async with asyncio.timeout(ROLE_PROPAGATION_TIMEOUT):
            while True:
                try:
                    await _run_medical_scribe_session(role_arn, s3_bucket)
                    return
                except BadRequestException as e:
                    last_error = e
                    await asyncio.sleep(ROLE_PROPAGATION_RETRY_DELAY)
    except TimeoutError:
        assert last_error is not None
        raise last_error
```

clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py
```diff
@@ -0,0 +1,148 @@
...
+                "Action": "sts:AssumeRole",
+            }
+        ],
+    }
+
+    try:
+        iam_client.create_role(
+            RoleName=role_name, AssumeRolePolicyDocument=json.dumps(trust_policy)
+        )
+    except iam_client.exceptions.EntityAlreadyExistsException:
```
alexgromero: Nit: now that `role_name` has a UUID suffix, `EntityAlreadyExistsException` shouldn't fire so it could be removed for clarity.

clients/aws-sdk-transcribe-streaming/tests/integration/conftest.py
```diff
@@ -0,0 +1,145 @@
...
+        iam_client.delete_role(RoleName=role_name)
+    except iam_client.exceptions.NoSuchEntityException:
+        pass
+
+
+@pytest.fixture(scope="session")
+def healthscribe_resources():
+    """Create HealthScribe resources for the test session and delete them after."""
+    unique_suffix = uuid.uuid4().hex
+    role_name = f"HealthScribeIntegTestRole-{unique_suffix}"
```
SamRemis: [non-blocking suggestion]

If this integ test is stopped mid-run (which can sometimes happen on our release infrastructure), then these resources will leak.  Over time, we risk running into limits on things like number of buckets, and just generally costing money in the account.

There are two things we can update to make it easier to detect and delete these resources:
1. We can prefix all new resources with something like `integ-test-` so that we can find the resources by prefix instead of a contains, even across services (so `integ-test-transcribe-streaming`, `integ-test-s3`, etc.)
2. We can add tags to any resources that support it (buckets and iam roles/trust policies both do), so we can detect these using [get_resources](https://docs.aws.amazon.com/boto3/latest/reference/services/resourcegroupstaggingapi/client/get_resources.html) or similar.

Alan4506: Both suggestions adopted.

alexgromero (approved):

PR #59, opened 2026-05-20
Retry on throttle and quota errors in integ test

Alan4506:
*Description of changes:*

The `test_get_medical_scribe_stream` integration test previously only retried on `BadRequestException` (IAM role propagation delay). When multiple test runs execute concurrently against the same account, the test can also encounter:

- `LimitExceededException`: post-stream analytics job quota exceeded
- `ThrottlingException` (surfaced as `CallError`): concurrent streaming session limit exceeded

This change adds retry handling for both cases. The `ThrottlingException` is not modeled in the service's error registry, so it arrives as a generic `CallError`, and we match it by error id in the message string.

*Testing:*

Without these changes, running 20 concurrent instances of integration tests:

```bash
for i in $(seq 1 20); do python -m pytest tests/integration/ -s & done; wait
```

results in failures with `LimitExceededException` and `CallError: Unknown error ... ThrottlingException`. With the fix applied, 20/20 concurrent runs pass.

### DIFF
--- a/clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
+++ b/clients/aws-sdk-transcribe-streaming/tests/integration/test_non_streaming.py
@@ -7,12 +7,15 @@ import asyncio
 import time
 import uuid
 
+from smithy_core.exceptions import CallError
+
 from aws_sdk_transcribe_streaming.models import (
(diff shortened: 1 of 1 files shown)


jonathan343 (approved):
LGTM!

PR #62, opened 2026-05-22
Add integration tests for connecthealth

Alan4506:
### Description of Changes

This PR adds integration tests for the Connect Health service client (`aws-sdk-connecthealth`), covering the following:

- Non-streaming operations: `create_domain`, `list_domains`, `get_domain`, `delete_domain`, `create_subscription`, `list_subscriptions`, `get_subscription`, `deactivate_subscription`, `get_medical_scribe_listening_session`.
- Bidirectional event streaming operation: `start_medical_scribe_listening_session`.

The test structure and patterns follow the same conventions established in the Lex Runtime V2 integration tests (PR #51), such as session-scoped conftest fixtures for resource setup/teardown.

### Testing

The integration tests create real AWS resources (Domain, Subscription, S3 bucket) under the calling AWS identity, run a live streaming session against the Connect Health service, then delete those resources. The calling identity therefore needs Admin (or equivalent) permissions on ConnectHealth and S3.

1. Assume an Admin role and export the resulting credentials so the new SDK (which uses `EnvironmentCredentialsResolver`) can pick them up:

   ```bash
   export AWS_ACCESS_KEY_ID=...
   export AWS_SECRET_ACCESS_KEY=...
   export AWS_SESSION_TOKEN=...
   ```

2. From `clients/aws-sdk-connecthealth/`, install the client and test dependencies:

   ```bash
   uv pip install -e . --group test
   ```

2. Run the tests

   ```bash
   python -m pytest tests/
   ```


clients/aws-sdk-connecthealth/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+
+def _create_s3_bucket(s3_client: Any, bucket_name: str) -> None:
+    """Create and tag the S3 bucket used by the streaming session.
+
+    Args:
+        s3_client: A boto3 S3 client.
+        bucket_name: The name of the S3 bucket to create.
+    """
+    s3_client.create_bucket(Bucket=bucket_name)
+    s3_client.put_bucket_tagging(
```
SamRemis: [nit]

Two separate http calls means we could create the bucket but fail to tag it.  We can include tags on the initial request via CreateBucketConfiguration.  See [the docs](https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/create_bucket.html) for the syntax.

clients/aws-sdk-connecthealth/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+
+    DeleteDomain rejects a Domain whose Subscription is not yet deactivated.
+
+    Args:
+        client: A ConnectHealth client.
+        domain_id: The parent Domain ID.
+        subscription_id: The Subscription ID to poll.
+    """
+    deadline = asyncio.get_event_loop().time() + _SUBSCRIPTION_POLL_TIMEOUT_SECONDS
+    while asyncio.get_event_loop().time() < deadline:
```
SamRemis: [nit/question]

Did we consider using [asyncio.timeout](https://docs.python.org/3/library/asyncio-task.html#asyncio.timeout) here?  Something like this:
```
          async with asyncio.timeout(_SUBSCRIPTION_POLL_TIMEOUT_SECONDS):
              while True:
                  response = await client.get_subscription(
                      input=GetSubscriptionInput(
                          domain_id=domain_id, subscription_id=subscription_id
                      )
                  )
                  if (
                      response.subscription is not None
                      and response.subscription.status == SubscriptionStatus.INACTIVE
                  ):
                      return
                  await asyncio.sleep(_SUBSCRIPTION_POLL_INTERVAL_SECONDS)
```

IMO that's a little more readable and effectively the same, but it's personal preference.  Though the above will cancel a request while it's in flight, and the existing code always finishes the given iteration.

clients/aws-sdk-connecthealth/tests/integration/test_bidirectional_streaming.py
```diff
@@ -0,0 +1,215 @@
...
+    got_transcript = results[1]
+    assert got_transcript, (
+        "Expected to receive a transcript event with non-empty content"
+    )
+
+    response = await client.get_medical_scribe_listening_session(
+        input=GetMedicalScribeListeningSessionInput(
+            session_id=session_id, domain_id=domain_id, subscription_id=subscription_id
+        ),
+        plugins=[streaming_plugin],
```
SamRemis: [_minor_ nit]
This confused me a bit since this isn't a streaming operation, but it still uses [the streaming endpoint](https://github.com/aws/api-models-aws/blob/main/models/connecthealth/service/2025-01-29/connecthealth-2025-01-29.json#L1438-L1473).  Maybe naming it `streaming_endpoint_plugin` would be easier to read, but this one is really not that big of a deal.

clients/aws-sdk-connecthealth/tests/integration/conftest.py
Alan4506: Thanks! Addressed in https://github.com/awslabs/aws-sdk-python/pull/62/changes/62eb957e684f6eef0a87f484ba47fa8291775651.

clients/aws-sdk-connecthealth/tests/integration/conftest.py
Alan4506: Thanks for the suggestion! I considered `asyncio.timeout`, but I'd lean towards keeping the current approach. As you noted, it always finishes the given iteration rather than cancelling it mid-request, which feels a bit safer.

That said, on a related [comment](https://github.com/awslabs/aws-sdk-python/pull/54#discussion_r3463943987) in another PR we did switch from `asyncio.get_event_loop()` to `asyncio.get_running_loop()`. So I also applied the switch for this PR in https://github.com/awslabs/aws-sdk-python/pull/62/changes/62eb957e684f6eef0a87f484ba47fa8291775651. In this case, it also keeps things consistent with the other existing tests.

clients/aws-sdk-connecthealth/tests/integration/test_bidirectional_streaming.py
Alan4506: Thanks, agreed there's some confusion here. However, naming it `streaming_endpoint_plugin` directly would shadow the imported function of the same name and cause an `UnboundLocalError` on the following line:
```
streaming_endpoint_plugin = streaming_endpoint_plugin(REGION)
```
 so I renamed it to `endpoint_plugin` instead to reduce the confusion. Done in [62eb957](https://github.com/awslabs/aws-sdk-python/pull/62/changes/62eb957e684f6eef0a87f484ba47fa8291775651).

SamRemis (approved):

PR #72, opened 2026-07-31
Add sts, imds, and http credential provider packages

arandito:
### Description

This PR introduces three new packages that register network credential providers into the SDK's modular AWS credential chain:

- **`aws-credentials-imds`** - EC2 Instance Metadata Service (IMDSv2) credential resolver and `Ec2InstanceMetadata` chain provider
- **`aws-credentials-http`** - container HTTP credential resolver (ECS/EKS) and `EcsContainer` chain provider
- **`aws-credentials-sts`** - STS AssumeRole resolvers and `ProfileAssumeRole` chain provider

**IMDS and HTTP are direct ports** of the existing `smithy_aws_core.identity.imds` and `container` resolvers, which will be deprecated. The credential resolution behavior is mostly identical. Changes are limited to:

- Module reorganization into standalone packages (`client.py`/`resolvers.py`/`providers.py` split).
- Naming cleanup (e.g. `ContainerMetadataClient` becomes `HttpCredentialsClient`, `EC2Metadata` becomes `IMDSClient`, etc).
- `ContainerCredentialsConfig` is flattened into class constructor arguments to make user interface cleaner.
- New provider/chain wiring that sources configuration from the environment and shared config profile. The original resolvers had no chain integration.
- Config-validation errors now raise package-specific `*ConfigurationError(SmithyError)` types instead of bare `ValueError`, reserving `SmithyIdentityError` for resolution-time failures.

**STS is new**. It ships two resolvers that separate the AssumeRole call itself from the profile configuration that feeds it:

- **`AssumeRoleCredentialsResolver`** performs the STS `AssumeRole` call. It takes an explicit `role_arn` and a `source_resolver` that provides the credentials used to make the call. This is the low-level resolver, with no knowledge of profiles, and can be used standalone outside the chain.
- **`ProfileAssumeRoleCredentialsResolver`** is the profile-driven resolver used by the chain. It reads a profile from the shared config file and resolves the credential source from that profile's `source_profile` (chaining to another profile, including nested role chains that terminate in static credentials) or `credential_source` (delegating to the `Environment`, `EcsContainer`, or `Ec2InstanceMetadata` provider). It then hands that source to an `AssumeRoleCredentialsResolver` to perform the call.

Splitting the two keeps the STS call logic isolated and reusable. `AssumeRoleCredentialsResolver` can be constructed directly with any source resolver, while `ProfileAssumeRoleCredentialsResolver` owns only the profile parsing and source resolution.

> [!IMPORTANT]
> The underlying STS client used for Assume Role calls in imported from the `aws-sdk-sts` client. This means that `aws-credentials-sts` has a required dependency on `aws-sdk-sts`. This does not cause a dependency cycle as the `aws-sdk-sts` client will never have a required dependency on `aws-credentials-sts` and instead is opt in. If artifact size for `aws-credentials-sts` becomes a concern due to the full import of the STS client package, we can explore a slim client implementation. For now, this is the most maintainable solution.

### Usage

Installing any of these packages auto-registers its provider into the SDK's credential chain via [entry points](https://packaging.python.org/en/latest/specifications/entry-points/). Each resolver can also be used directly:

```python
from aws_credentials_imds import IMDSCredentialsResolver
from aws_credentials_http import ContainerCredentialsResolver
from aws_credentials_sts import AssumeRoleCredentialsResolver

# IMDS / container: HTTP client defaults to AIOHTTPClient
resolver = IMDSCredentialsResolver()
credentials = await resolver.get_identity(properties={})

# STS AssumeRole with any source resolver
resolver = AssumeRoleCredentialsResolver(
    source_resolver=source_resolver,
    role_arn="arn:aws:iam::123456789012:role/example",
)
credentials = await resolver.get_identity(properties={})
```

### Testing
- Ported existing test suites for IMDS and Container, with slight modification for new interfaces.
- Added unit test coverage for new STS credential resolvers and for the three new chain providers.
- Tested all three packages standalone and integrated in the new `IdentityChain` with live service calls
  - Configuration:
    - STS: Set up iam roles to assume on AWS Console
    - IMDS: Set up EC2 instance with an IMDS instance profile + role.
    - HTTP: Used [amazon-ecs-local-container-endpoints](https://github.com/awslabs/amazon-ecs-local-container-endpoints) to host container endpoints locally.
  - Confirmed all providers are configured automatically with `~/.aws/config` profiles and environment variables.
  - Confirmed chain order is preserved when multiple sources are configured.
  - Confirmed IMDS and HTTP work as credential sources for STS providers


packages/aws-credentials-http/src/aws_credentials_http/client.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(
+        self,
+        http_client: HTTPClient,
+        *,
+        timeout: int = _DEFAULT_TIMEOUT,
+        retries: int = _DEFAULT_RETRIES,
+    ):
+        self._http_client = http_client
+        self._timeout = timeout
```
Alan4506, line 39: `self._timeout` seems never used. Should it be applied to the request, or should the parameter be dropped? Or do you want to to add a TODO?

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
```diff
@@ -0,0 +1,134 @@
...
+                auth_token = await asyncio.to_thread(self._read_file, filename)
+            except (FileNotFoundError, PermissionError) as error:
+                raise SmithyIdentityError(
+                    f"Unable to open {os.environ[self.ENV_VAR_AUTH_TOKEN_FILE]}."
+                ) from error
+
+            fields.set_field(Field(name="Authorization", values=[auth_token]))
+        elif self.ENV_VAR_AUTH_TOKEN in os.environ:
+            auth_token = os.environ[self.ENV_VAR_AUTH_TOKEN]
+            fields.set_field(Field(name="Authorization", values=[auth_token]))
```
Alan4506: The token goes into the `Authorization` header unvalidated, while [botocore](https://github.com/boto/botocore/blob/21bce1911c4af5a1ad9eda564e3d823770a89017/botocore/credentials.py#L2107-L2121) rejects `\r` and `\n`: 
```python
def _build_headers(self):
    auth_token = None
    if self.ENV_VAR_AUTH_TOKEN_FILE in self._environ:
        auth_token_file_path = self._environ[self.ENV_VAR_AUTH_TOKEN_FILE]
        with open(auth_token_file_path) as token_file:
            auth_token = token_file.read()
    elif self.ENV_VAR_AUTH_TOKEN in self._environ:
        auth_token = self._environ[self.ENV_VAR_AUTH_TOKEN]
    if auth_token is not None:
        self._validate_auth_token(auth_token)
        return {'Authorization': auth_token}

def _validate_auth_token(self, auth_token):
    if "\r" in auth_token or "\n" in auth_token:
        raise ValueError("Auth token value is not a legal header value")
```
Should we add some checks as well?

packages/aws-credentials-http/src/aws_credentials_http/client.py
```diff
@@ -0,0 +1,93 @@
...
+            except Exception as error:
+                last_exc = error
+                await asyncio.sleep(_SLEEP_SECONDS)
+                attempts += 1
+
+        raise SmithyIdentityError(
+            f"Failed to retrieve container metadata after {self._retries} attempt(s)"
+        ) from last_exc
+
+    def _validate_allowed_url(self, uri: URI) -> None:
```
Alan4506, line 81: [botocore](https://github.com/boto/botocore/blob/21bce1911c4af5a1ad9eda564e3d823770a89017/botocore/utils.py#L3100-L3112) allows any host over HTTPS and only restricts plain HTTP to loopback/allowlisted hosts:
```python
def _validate_allowed_url(self, full_url):
    parsed = botocore.compat.urlparse(full_url)

    if parsed.scheme == 'https':
        return
    if self._is_loopback_address(parsed.hostname):
        return
    is_whitelisted_host = self._check_if_whitelisted_host(parsed.hostname)
    if not is_whitelisted_host:
        raise ValueError(
            f"Unsupported host '{parsed.hostname}'.  Can only retrieve metadata "
            f"from a loopback address or one of these hosts: {', '.join(self._ALLOWED_HOSTS)}"
        )

```
Are we missing the HTTPS check here, or is it intentional?

alexgromero: Agreed, this seems to be a miss from the initial implementation. Per the credential provider spec, HTTPS endpoints should be allowed regardless of host (only plain HTTP is restricted to loopback/link-local), so this should be checked here.

packages/aws-credentials-http/tests/unit/test_client.py
```diff
@@ -0,0 +1,108 @@
...
+    _assert_expected_credentials(credentials, "akid123", "s3cr3t", "session_token")
+
+
+async def test_client_https_host() -> None:
+    response_body = json.dumps(DEFAULT_RESPONSE_DATA)
+    http_client = mock_http_client_response(200, response_body.encode())
+    client = HttpCredentialsClient(http_client)
+
+    credentials = await client.get_credentials(
+        URI(scheme="https", host="169.254.170.2"), Fields()
```
alexgromero: This test uses `169.254.170.2` which is allowlisted so it passes whether or not the scheme is checked.

We should update it to use a non-allowlisted host like `URI(scheme="https", host="example.com")` that way it actually covers the HTTPS case.

alexgromero (changes requested):
One callout across all three new credential packages: they add async tests, but the package local `pyproject.toml` files don't declare the test setup needed to run them standalone.

This worked in `smithy-python` because it already had that wiring, but here `uv sync && uv run pytest` does not install an async pytest plugin. Even with `pytest-asyncio`, the async tests still need either `@pytest.mark.asyncio` or `asyncio_mode = "auto"`.

Could we add a shared test dependency group and pytest async configuration to each package so their test suites can run on their own?

packages/aws-credentials-imds/src/aws_credentials_imds/client.py
```diff
@@ -0,0 +1,178 @@
...
+                    scheme=self._base_uri.scheme,
+                    host=self._base_uri.host,
+                    port=self._base_uri.port,
+                    path=self._TOKEN_PATH,
+                ),
+                fields=headers,
+            )
+            response = await self._http_client.send(request)
+            token_value = await response.consume_body_async()
+            self._token = IMDSToken(token_value.decode("utf-8"), self._config.token_ttl)
```
alexgromero, line 140: Should we fail on a non-200 response? It seems a 4xx response would get stored as the token value here.

packages/aws-credentials-sts/src/aws_credentials_sts/resolvers.py
```diff
@@ -0,0 +1,367 @@
...
+                    region=self._region,
+                    transport=self._http_client,
+                )
+            )
+
+        response = await self._client.assume_role(
+            AssumeRoleInput(
+                role_arn=self._role_arn,
+                role_session_name=self._role_session_name,
+                external_id=self._external_id,
```
alexgromero, line 145: botocore also reads `duration_seconds` from the profile and passes it as `DurationSeconds`. Without it, assumed sessions always fall back to the STS default (1h), so a profile setting `duration_seconds` is silently ignored. Do you think it's worth adding for parity or is it intentionally out of scope for now?

arandito: It was intentionally out of scope as its not required in the internal spec. However, it's was very cheap to add support so its added in 524a395.

Other options such as `mfa_serial` are a bit more complicated since they require some sort of callback or prompting design. A lot of newer SDKs don't support it through a profile so I am leaving it out for now. We can always add it later if there is demand.

packages/aws-credentials-imds/src/aws_credentials_imds/client.py
arandito: Yes, that was an existing gap. We now raise a `SmithyIdentityError` for both call sites if the response is not a 2xx, similar to the container provider. When we implement retries for IMDS, we will need more granular handling for retryable status codes, but that will be future work.

arandito:
@alexgromero 
> Could we add a shared test dependency group and pytest async configuration to each package so their test suites can run on their own

I'm actually going to follow up with a PR that introduces CI + testing infrastructure using uv workspaces. This is similar to Smithy Python's setup. We will need this for our internal infrastructure to test these packages as part of our release process. For now, I'd prefer we keep the pytest configurations and dependencies centralized, instead of per-package, to avoid drift.

packages/aws-credentials-http/tests/unit/test_client.py
arandito: Updated in [524a395](https://github.com/aws/aws-sdk-python/pull/72/commits/524a395adac33048707e1abb93a4ae005aa57099).

packages/aws-credentials-http/src/aws_credentials_http/client.py
arandito: Good catch! Addressed in [524a395](https://github.com/aws/aws-sdk-python/pull/72/commits/524a395adac33048707e1abb93a4ae005aa57099).

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
arandito: Yes! Addressed in [524a395](https://github.com/aws/aws-sdk-python/pull/72/commits/524a395adac33048707e1abb93a4ae005aa57099).

packages/aws-credentials-http/src/aws_credentials_http/client.py
arandito: I'm not sure why this wasn't applied to the request in the first place.

I confirmed we should apply the default timeout by looking at botocore. However, we need to apply the value to both the read timeout and connect timeout. Looks like `HTTPRequestConfiguration` only supports `read_timeout` for now which may be why it wasn't used.

I applied the value as the `read_timeout` in the request configuration and added a TODO to apply the value to `connect_timeout` once its supported.

Alan4506:
The new commit LGTM overall. 

Regarding https://github.com/aws/aws-sdk-python/pull/72#discussion_r3727694222, we should have a follow-up though: in [smithy-http](https://github.com/smithy-lang/smithy-python/blob/5c25af99413f0ff404437e66da1aaaeb73ae0347/packages/smithy-http/src/smithy_http/aio/aiohttp.py#L76-L112), `AIOHTTPClient.send` currently accepts `request_config` but never consumes it. We'll need follow-ups in smithy-http to actually wire `read_timeout` into the request and also add `connect_timeout` support.

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
```diff
@@ -0,0 +1,139 @@
...
+            secret_access_key=secret_access_key,
+            session_token=session_token,
+            expiration=expiration,
+            account_id=account_id,
+        )
+        return self._credentials
+
+    async def invalidate(self) -> None:
+        """Discard cached credentials so the next resolution re-queries the endpoint."""
+        self._credentials = None
```
alexgromero: Should we make this a no-op for now? Clearing the cached credentials works against static stability. The refresh behavior we're moving toward expires them only when the rejected identity matches, rather than discarding them outright.

Since we can't implement the identity-match behavior yet (it needs the rejected identity passed in), a no-op seems safer than carrying over the clear. What do you think?

arandito: Good call. Added in [7262e98](https://github.com/aws/aws-sdk-python/pull/72/commits/7262e9849bfb5ead526bb02d6ef687e75a606f62).

These can be updated once we update the `IdentityResovler.invalidate()` protocol in `smithy-http`.

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
```diff
@@ -0,0 +1,135 @@
...
+                path=os.environ[self.ENV_VAR],
+            )
+        elif self.ENV_VAR_FULL in os.environ:
+            parsed = urlparse(os.environ[self.ENV_VAR_FULL])
+            return URI(
+                scheme=parsed.scheme,
+                host=parsed.hostname or "",
+                port=parsed.port,
+                path=parsed.path,
+            )
```
alexgromero, line 99: Also a gap from the initial implementation: the full-URI branch rebuilds the URI but doesn't carry `parsed.query` over, so query strings get dropped. The provider spec says the full URI should be taken as-is and has a test case with `?foo=bar%20baz`. Can we add `query=parsed.query` here?

packages/aws-credentials-http/README.md
jonathan343, line 1: I think the README's can be improved by providing example of how to use these packages with the AWS SDK for Python as well as standalone usage. 

Not gonna block approval on this, but something we should consider now or in a follow up.

packages/aws-credentials-imds/src/aws_credentials_imds/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+__version__ = "0.0.0"
+
+from .client import IMDSConfigurationError
+from .providers import Ec2InstanceMetadataProvider
+from .resolvers import IMDSCredentialsResolver
+
+__all__ = (
```
jonathan343, line 9: We should probably export `IMDSConfig` here as well

packages/aws-credentials-sts/src/aws_credentials_sts/resolvers.py
```diff
@@ -0,0 +1,370 @@
...
+_ROLE_ARN = "role_arn"
+_ROLE_SESSION_NAME = "role_session_name"
+_EXTERNAL_ID = "external_id"
+_DURATION_SECONDS = "duration_seconds"
+_SOURCE_PROFILE = "source_profile"
+_CREDENTIAL_SOURCE = "credential_source"
+_REGION = "region"
+
+_CREDENTIAL_SOURCE_SLOTS = {
+    "Environment": StandardProvider.ENVIRONMENT,
```
jonathan343, line 48: For my own knowledge, how does the chain know when set environment variables are to be used directly or when to be used as the source for an assume role resolver?

jonathan343 (changes requested):
I have some very minor feedback, but overall looks good!

packages/aws-credentials-sts/README.md
```diff
@@ -0,0 +1,5 @@
+# aws-credentials-sts
+
+This package provides STS-based credential resolvers and chain providers. 
+Installing it automatically adds the `PROFILE_ASSUME_ROLE` source to the SDK's
```
jonathan343: What is the `PROFILE_ASSUME_ROLE` name referencing?

packages/aws-credentials-sts/src/aws_credentials_sts/resolvers.py
arandito: After offline discussion, the chain will ignore the environment credentials provider if a user has explicitly set a profile that configures assume role with an environment credential source.

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
arandito: Added in [e0b561f](https://github.com/aws/aws-sdk-python/pull/72/commits/e0b561fb74829d65e82f26bdc6f5a739a3ef3564).

packages/aws-credentials-http/README.md
arandito: Updated all READMEs to provide this information.

packages/aws-credentials-imds/src/aws_credentials_imds/__init__.py
arandito: Added in [e0b561f](https://github.com/aws/aws-sdk-python/pull/72/commits/e0b561fb74829d65e82f26bdc6f5a739a3ef3564).

packages/aws-credentials-sts/README.md
arandito: Rewrote all READMEs and removed this wording to make it more clear for customers.

jonathan343 (approved):

packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
alexgromero: Thanks! Not going to block on this, but as a follow-up it'd be good to add a test with a query string.

alexgromero (approved):

PR #79, opened 2026-08-16
Update STS client to use new async config and fix failing container resolver tests

arandito:
### Description

This PR makes two small runtime package updates:

#### `aws-credentials-sts`: adopts new async config API
Updates `AssumeRoleCredentialsResolver` to build its `AsyncSTSClient` from `AsyncSTSConfig.resolve(...)` instead of the deprecated synchronous `Config`. Config overrides (source credentials resolver, region, and optional transport) are now assembled into a dict and resolved asynchronously, with `transport` only set when an HTTP client is provided.

> [!WARNING]
>  This depends on the parallel `smithy-python` change that deprecates the old `Config` in favor of `AsyncSTSConfig` (https://github.com/smithy-lang/smithy-python/pull/758). It **must be sim-shipped in the same release** as that PR; releasing independently will break `aws-credentials-sts` (`ImportError: cannot import name 'AsyncSTSConfig'`).

#### `aws-credentials-http`: normalize empty query strings
Fixes `ContainerCredentialsResolver` to pass `query=parsed.query or None` when resolving from `AWS_CONTAINER_CREDENTIALS_FULL_URI`. `urlparse` yields `''` for a query-less URL, and the current `URI` no longer normalizes that to `None`, so the constructed request URI didn't match expectations.

### Testing

Ran the container credentials resolver unit tests standalone:
```shell
cd packages/aws-credentials-http
uv run --no-project --python 3.12 \
  --with pytest --with pytest-asyncio \
  --with-editable . \
  python -m pytest tests/unit/test_resolvers.py -o asyncio_mode=auto
```
All 12 tests pass, including the four `test_resolver_env_*` cases that previously failed on the empty-query-string mismatch.

9 `aws-credentials-sts` resolver tests currently fail with `ImportError: cannot import name 'AsyncSTSConfig' from 'aws_sdk_sts.config'` until smithy-python#758 ships the new async config. Nothing to do here until that lands; they'll pass once the two are released together.

### DIFF
--- a/packages/aws-credentials-http/.changes/next-release/aws-credentials-http-bugfix-bd4ef625870a44faa848494a7489ad19.json
+++ b/packages/aws-credentials-http/.changes/next-release/aws-credentials-http-bugfix-bd4ef625870a44faa848494a7489ad19.json
@@ -1,0 +1,4 @@
+{
+  "type": "bugfix",
+  "description": "Normalize empty container credential URI query strings to `None`."
+}
--- a/packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
+++ b/packages/aws-credentials-http/src/aws_credentials_http/resolvers.py
@@ -95,7 +95,7 @@ class ContainerCredentialsResolver(
                 host=parsed.hostname or "",
                 port=parsed.port,
                 path=parsed.path,
-                query=parsed.query,
+                query=parsed.query or None,
             )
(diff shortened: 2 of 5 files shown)


Alan4506 (dismissed):
LGTM! Thanks @arandito

alexgromero (approved):

PR #82, opened 2026-08-19
Use async configs in integ tests and examples

Alan4506:
This PR updates all integration tests and examples to use the new async-resolved config classes instead of the deprecated `Config` constructor:

```python
config=await AsyncPollyConfig.resolve(...)  # was: config=Config(...)
```

Client-creation helpers are now `async` and all call sites `await` them.


alexgromero (approved):
LGTM :shipit:

PR #85, opened 2026-08-20
Update README for Developer Preview

Alan4506:
*Description of changes:*

Updates the root README for the Developer Preview:

- Add a preview status callout recommending Boto3 for production, a key features list, and an SDK vs. Boto3 comparison table.
- Add installation instructions (including how to install CRT as extra).
- Add a quick start.
- Add a Resources section (SDK homepage, developer guide, API reference).

**Note:** content reflects the final dev preview release state. Changes such as https://github.com/smithy-lang/smithy-python/pull/765 has not been released yet. The developer guide has not been published yet as well. We should merge this PR only after all those changes land. At that time we'll update the "About" section of the repo as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,6 @@ aws_secret_access_key = YOUR_SECRET_ACCESS_KEY
 aws_session_token = YOUR_SESSION_TOKEN
 ```
 
-Omit `aws_session_token` if you use long-term credentials.
 For the credential sources that the Developer Preview supports, see
 [Credential providers](https://docs.aws.amazon.com/sdk-for-python/v1/guide/credential-providers.html)
 in the developer guide.
--- a/README.md
+++ b/README.md
@@ -1,31 +1,133 @@
-## AWS SDK for Python
+## AWS SDK for Python (Developer Preview)
 [![Apache 2 licensed][apache-badge]][apache-url]
 
 [apache-badge]: https://img.shields.io/badge/license-APACHE2-blue.svg
 [apache-url]: https://github.com/aws/aws-sdk-python/blob/main/LICENSE
 
-This repository contains experimental async clients for the AWS SDK for Python.
-These new clients will allow you to interact with select AWS services that can
-best utilize Python's async functionality. Unlike Boto3, these clients are
-distributed per-service, leaving you the option to pick what fits your needs.
+> **Developer Preview — Not for production use**
+>
+> This SDK is in **Developer Preview** and is intended for evaluation and testing in pre-production environments only. Do not use it for production workloads. APIs and behavior might change before general availability.
+>
+> **For production deployments, use [Boto3](https://github.com/boto/boto3)** — the established, generally available AWS SDK for Python with full coverage of all AWS services.
 
-Please note that this new project is in early development and will be seeing
-rapid iteration over the coming months. This may mean instability in both
-public interfaces and general behaviors. Until the project releases version
-1.0.0, breaking changes may occur between minor versions of the SDK. We'd
-strongly advise strict pinning to a version of the SDK for any non-experimental
-use cases.
+---
 
-While we're developing, we welcome all feedback. Please feel free to note things
-you like, dislike, or would like to see in a future release. We'll be using that
-to help drive the direction of the product going forward.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,31 +1,133 @@
...
+
+The meta-package installs only the service clients that you select as extras;
+without extras, it installs no clients. It coordinates client versions through
+its MAJOR.MINOR version, so it installs compatible client versions together.
+
+Use the meta-package when your application depends on several clients. Install
+individual client packages when your application uses only one or two services
+and you want the smallest dependency set.
+
+### Bidirectional streaming transport
```
alexgromero: I would move this after Quick start so Installation flows directly into the basic example.

README.md
```diff
@@ -1,31 +1,133 @@
...
+
+**Choose [Boto3](https://github.com/boto/boto3) if you** need production stability, full AWS service coverage, synchronous workloads, or features like paginators, waiters, and presigned URLs.
+
+**Choose this SDK if you** are evaluating native async capabilities, building high-throughput async prototypes, working with streaming services in test environments, or want to provide feedback to shape the GA release.
+
+## Installation
+
+The SDK requires Python 3.12 or later and publishes a separate package for
+each AWS service client. Install the client packages that the following
+examples use:
```
alexgromero: I like showing the different ways to install clients here, but I wonder if we need to frame it around the examples below. Could this section just explain client installation generally, then Quick start installs `aws-sdk-sts` for its example? We could also use different clients here for a little more variety, instead of installing the same Bedrock client again below with the `awscrt` extra.

README.md
```diff
@@ -1,31 +1,133 @@
...
+
+```
+python -m pip install aws-sdk-sts aws-sdk-bedrock-runtime
+```
```
alexgromero: small nit: could we use bash for the install commands? GitHub gives them a bit of syntax highlighting.

README.md
```diff
@@ -1,31 +1,133 @@
...
+
+async def create_client() -> AsyncBedrockRuntimeClient:
+    config = await AsyncBedrockRuntimeConfig.resolve(
+        region="us-east-1",
+        transport=AWSCRTHTTPClient(),
+    )
+    return AsyncBedrockRuntimeClient(config=config)
+```
+
+## Quick start
```
alexgromero: Could we make this a little more step-by-step? Like install aws-sdk-sts, make sure region/credentials are configured, then run the example. Similar to [Using Boto3](https://github.com/boto/boto3#using-boto3).

alexgromero (commented):
Overall this looks good! I think we could tweak the flow a bit to get users to the Quick start sooner. It might read more naturally as intro --> installation --> Quick start --> additional setup/details like bidirectional streaming.

README.md
```diff
@@ -1,31 +1,159 @@
...
+[default]
+aws_access_key_id = YOUR_ACCESS_KEY_ID
+aws_secret_access_key = YOUR_SECRET_ACCESS_KEY
+aws_session_token = YOUR_SESSION_TOKEN
+```
 
-The SDK uses **GitHub Issues** to track feature requests and issues with the SDK. In addition,
-we intend to use **GitHub Projects** to provide a high level overview of our roadmap and the
-features we are actively working on.
+Omit `aws_session_token` if you use long-term credentials.
```
alexgromero: Do we want to call out long-term credentials here? Since this is the Quick start, I wonder if we should keep the example focused on temporary credentials and point to the credential providers guide for other supported options.

README.md
```diff
@@ -1,31 +1,159 @@
...
+    async with AsyncSTSClient() as client:
+        response = await client.get_caller_identity(GetCallerIdentityInput())
+        print(f"Account: {response.account}")
+        print(f"Arn: {response.arn}")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
```
alexgromero: Maybe we can add a short note like "See the [STS API Reference](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/sts/) for available operations and models."

alexgromero (approved):

PR #87, opened 2026-08-21
Improve the API reference landing page

jonathan343:
## Overview

The API reference currently inherits its landing page from the repository README, which leaves readers without a clear introduction to the SDK or path into the reference documentation.

This PR replaces it with a purpose-built overview and adds a dismissible Developer Preview notice across the docs. The banner ensures that users entering through deep links still see the production-use guidance.

## Testing / Preview

- `make docs-install`
- `make docs`
- `make docs-serve`

Screenshot 2026-08-21 at 12 08 01 AM

---

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3,7 +3,7 @@
 The AWS SDK for Python provides asynchronous clients for supported AWS services.
 Each service has its own package.
 
-!!! warning "Developer preview: Not for production use"
+!!! warning "Developer Preview: Not for production use"
 
     Use this SDK for evaluation and pre-production testing. Interfaces and
     behavior may change before general availability. Use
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -1,3 +1,88 @@
...
+
+
+asyncio.run(main())
+```
+
+`AsyncDynamoDBConfig.resolve()` loads credentials and other shared settings
+from the standard AWS configuration sources. The `region` argument overrides
+the configured region for this client.
+
+## Documentation and support
```
Alan4506, line 110: Not blocking:

We can add a link to the developer guide once it gets released in the future.

docs/index.md
```diff
@@ -1,3 +1,88 @@
...
+import asyncio
+
+from aws_sdk_dynamodb.client import AsyncDynamoDBClient
+from aws_sdk_dynamodb.config import AsyncDynamoDBConfig
+from aws_sdk_dynamodb.models import ListTablesInput
+
+
+async def main():
+    config = await AsyncDynamoDBConfig.resolve(region="us-east-1")
+    client = AsyncDynamoDBClient(config=config)
```
Alan4506: Let's modify this code block to sync up with the new pattern established in https://github.com/smithy-lang/smithy-python/pull/765. 

Consider `aysnc with` to use the context manager.

docs/index.md
```diff
@@ -1,3 +1,88 @@
...
+
+    ```bash
+    pip install "aws-sdk-python[bedrock_runtime,sts]"
+    ```
+
+The [available clients](clients/index.md) page lists every service package.
+
+## List DynamoDB tables
+
+After you [configure AWS credentials](https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html)
```
Alan4506: The link points to the sdkref Standardized credential providers page, which lists the full set of providers - but our developer preview only supports a subset. Readers may pick an unsupported method (e.g., SSO) and hit failures. 

Should we give users more explicit instructions about how to set up credentials, like https://github.com/aws/aws-sdk-python/pull/85? Once the developer guide gets released, we can link to one of its sections that lists out supported providers.

docs/index.md
```diff
@@ -1,3 +1,88 @@
...
+- **Streaming support.** Clients for streaming services support event streams
+  and bidirectional streaming.
+
+The code generator builds each client from its service's
+[Smithy](https://smithy.io/) model.
+
+## Install a client
+
+The clients require Python 3.12 or later.
+
```
Alan4506, line 40: Do we want to introduce virtual environment setup between Python version requirement and client installation?

See Boto3 documentation: https://docs.aws.amazon.com/boto3/latest/guide/quickstart.html.

Alan4506 (changes requested):
Thanks @jonathan343! Left some comments. Let me know if you have questions.

docs/index.md
jonathan343: Yes. If it's up before our nest release, I can update it in this PR. Otherwise, we can get it added in a follow up.

docs/index.md
jonathan343: Yup, I was waiting until it was merged which it is now. Added in 972c2efecb595d41d72ac30952fe61f2151eb14d

docs/index.md
jonathan343: Added in 972c2efecb595d41d72ac30952fe61f2151eb14d

docs/index.md
jonathan343: Added in 972c2efecb595d41d72ac30952fe61f2151eb14d

docs/index.md
```diff
@@ -1,3 +1,114 @@
...
+from aws_sdk_dynamodb.config import AsyncDynamoDBConfig
+from aws_sdk_dynamodb.models import ListTablesInput
+
+
+async def main():
+    config = await AsyncDynamoDBConfig.resolve(region="us-east-1")
+
+    async with AsyncDynamoDBClient(config=config) as client:
+        response = await client.list_tables(input=ListTablesInput(limit=10))
+        for table in response.table_names:
```
Alan4506: It looks like the `or []` was dropped from the line 98 in the latest commit. Without this, pyright now reports `error: Object of type "None" cannot be used as iterable value (reportOptionalIterable)`, since we have `table_names: list[str] | None = None` defined in [ListTablesOutput](https://github.com/aws/aws-sdk-python/blob/6952827f5944a646025eb441426eb3db615e60fd/clients/aws-sdk-dynamodb/src/aws_sdk_dynamodb/models.py#L17441).

Do we want to change it back to `for table in response.table_names or []:`?

jonathan343: I had removed it since in reality, DynamoDB sends `[]` when you have no tables in a specific region. I tested this with my dev account and verified `[]` was returned. So it felt weird to keep it, but it is modeled as optional and that's reflected in our codegen. I added it back in 567b9bf08c3f47ab6bd88c57c08a21a8b21ccd89

Alan4506 (approved):

adititiwari2202-ui (approved):

PR #93, opened 2026-09-10
Rename to AWS SDK for Python v4

Alan4506:
*Description of changes:*

This PR changes some occurrences of "AWS SDK for Python" to "AWS SDK for Python v4". There will be follow-up PRs to update the root README, the docs of each client, and some URLs from /v1 to /v4.


clients/aws-sdk-python/README.md
```diff
@@ -31,4 +31,4 @@ python -m pip install "aws-sdk-python[all]"
 - [API Reference](https://docs.aws.amazon.com/sdk-for-python/v1/reference/clients/) -
```
arandito, line 31: Should we move the api reference path to v4? Or are we gonna follow up with this once it exists?

clients/aws-sdk-python/README.md
```diff
@@ -31,4 +31,4 @@ python -m pip install "aws-sdk-python[all]"
...
   the available clients, operations, and types.
 - [Developer Guide](https://docs.aws.amazon.com/sdk-for-python/v1/guide/) -
```
arandito, line 33: Same question here, are we following up to update the path to v4 once it exists?

clients/aws-sdk-python/README.md
Alan4506: Yes, this is noted in the PR description. We'll update this once the new URL (v4) becomes valid.

clients/aws-sdk-python/README.md
Alan4506: Same as above. As noted in the PR description, we'll update this once the new URL (v4) becomes valid.

arandito (approved):
lgtm

PR #98, opened 2026-09-17
Add GitHub issue templates for bugs, features, and documentation

jonathan343:
## Summary

This PR adds bug report, feature request, and documentation forms under `.github/ISSUE_TEMPLATE/`. These templates can be seen consistently across repos like `boto/botocore`, `boto/boto3`, and `aws/aws-sdk-java-v2`. Each form carries a required affected-area dropdown (service client, credentials, installation, codegen, docs), and the bug and feature forms take a free-text field for affected package names and versions. The bug form also requires Python version and OS. `config.yml` disables blank issues and points open-ended questions at the Q&A and Ideas discussions, plus the Boto3 tracker for reports that belong there.
that belong there.

---


.github/ISSUE_TEMPLATE/documentation.yml
```diff
@@ -0,0 +1,37 @@
...
+      required: true
+
+  - type: dropdown
+    id: surface
+    attributes:
+      label: Which documentation?
+      description: Select all that apply.
+      multiple: true
+      options:
+        - API Reference (https://docs.aws.amazon.com/sdk-for-python/v1/reference/)
```
Alan4506, line 32: Just a note that we need to update this to v4 later, when the new v4 URL goes live.

.github/ISSUE_TEMPLATE/documentation.yml
```diff
@@ -0,0 +1,37 @@
...
+    validations:
+      required: true
+
+  - type: dropdown
+    id: surface
+    attributes:
+      label: Which documentation?
+      description: Select all that apply.
+      multiple: true
+      options:
```
Alan4506, line 31: The template description says: "Report an issue in the API Reference documentation or Developer Guide", but the Developer Guide is missing from the options here. Do we want to add it here?

Alan4506 (commented):
Thanks @jonathan343! Looks good overall - just 2 comments below.

.github/ISSUE_TEMPLATE/documentation.yml
jonathan343: Yea, since it's not live yet, I went with `v1`. Also assuming we will have redirects implemented so this should be safe

.github/ISSUE_TEMPLATE/documentation.yml
jonathan343: Thanks! Added in 82c66b6

Alan4506 (approved):

